PR #2, opened 2021-12-27
Vagrant host-only network IP changed to meet new VirtualBox rules

evanhsu:
# What
This PR changes the Vagrant host-only network IP from `192.168.13.37` to `192.168.56.37` to comply with the IP range restrictions introduced in VirtualBox v1.6.30

(The PR also adds the Snyk local cache file to the `.gitignore` in a totally unrelated commit 😬 ) 

# Why
VirtualBox introduced new restrictions on the IP range for host-only networks in version 1.6.30

Here's the VirtualBox docs about the IP range restriction:
https://www.virtualbox.org/manual/ch06.html#network_hostonly

> On Linux, Mac OS X and Solaris Oracle VM VirtualBox will only allow IP addresses in 192.168.56.0/21 range to be assigned to host-only adapters

# Some console output
Without this change in place, I get the following console output while running `vagrant up` (with VirtualBox v1.6.30):

```
vagrant up
Bringing machine 'test-cli-setup' up with 'virtualbox' provider...
==> test-cli-setup: Checking if box 'bento/ubuntu-20.04' version '202107.28.0' is up to date...
==> test-cli-setup: Clearing any previously set network interfaces...
There was an error while executing `VBoxManage`, a CLI used by Vagrant
for controlling VirtualBox. The command and stderr is shown below.

Command: ["hostonlyif", "ipconfig", "vboxnet0", "--ip", "192.168.13.1", "--netmask", "255.255.255.0"]

Stderr: VBoxManage: error: Code E_ACCESSDENIED (0x80070005) - Access denied (extended info not available)
VBoxManage: error: Context: "EnableStaticIPConfig(Bstr(pszIp).raw(), Bstr(pszNetmask).raw())" at line 242 of file VBoxManageHostonly.cpp
```

PR #3, opened 2021-12-28
QUICK-170 add notifications services to docker compose

evanhsu:
# WAT?
This PR adds two new services to the `docker-compose.yml` file to support the event-based Notifications system.
The new services are:
  - `notification-engine`
  - `notification-sender`

# Testing it

The setup steps that I added to the  README document the steps that I was using to set up a new Plextrac instance inside the Vagrant box.

  - Follow the setup instructions that have been added to the Readme to get a PlexTrac instance set up with this new docker-compose config.
  - Check the "Health" of the notification-engine and notification-sender (as reported by docker-compose via the new healthcheck) by running these commands inside the Vagrant box:
  
         docker-compose ps

    This should show that both the 'engine' and 'sender' are "Healthy".

  - Stop the Redis container. This should cause the notification services to become unhealthy (after a short delay).
  
         docker-compose stop redis

  - Continue to check the health of the services for about 40 seconds to make sure that docker-compose recognizes that they're unhealthy
  - Turn Redis back on and make sure that the services become healthy again automatically
  
         docker-compose up -d redis

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,15 @@ This is the PlexTrac management CLI that can be used to perform the initial setu
         configure                            does initial configuration required for PlexTrac application
         info                                 display information about the current PlexTrac Instance
 
-### For a fresh install of a new PlexTrac instance:
+## For a fresh install of a new PlexTrac instance:
 
     DOCKER_HUB_KEY=${DOCKER_HUB_KEY} plextrac configure
     plextrac update
     plextrac start
+
+### A quick note about the `DOCKER_HUB_KEY`
+It should look several groups of alphanumeric characters separated by hyphens. It should _not_ be base64-encoded.
+Example: `abcd123-11a1-22bb-c3c3-defg567890`
 
 ## Testing inside a Vagrant box
 
@@ -38,13 +42,15 @@ You'll need to have [Vagrant](https://www.vagrantup.com/) installed before you c
 
 3.  Create a default `.env` file by running the `configure` command
 
-        plextrac configure
+        DOCKER_HUB_KEY=<YOUR_DOCKER_HUB_KEY> plextrac configure
 
-4.  Set your DOCKER_HUB_KEY environment variable in the `.env` file
+    This will write a `.env` file to `/opt/plextrac/.env` with default values plus the DOCKER_HUB_KEY that you just provided.
+
+4.  Set any non-default environment variables in the `.env` file
 
         vim /opt/plextrac/.env
 
-    You should update the following vars:
+    You should check the following vars:
 
         DOCKER_HUB_KEY    # Your token for pulling images from the plextrac private Docker Hub registry
         UPGRADE_STRATEGY  # A docker image tag to pull (i.e. 'stable', 'edge', 'imminent', or a git commitish)
--- a/README.md
+++ b/README.md
@@ -1,1 +1,72 @@
 # plextrac-manager-util
+
+This is the PlexTrac management CLI that can be used to perform the initial setup of a new PlexTrac installation as well as performing maintenance tasks on running instances.
+
+## Usage
+
+    Usage: plextrac COMMAND [FLAGS]
+
+    [+] Available commands:
+        initialize                           initialize local system for PlexTrac installation
+        install                              install PlexTrac (assumes previously initialized system)
+        start                                run the PlexTrac application
+        backup                               perform backup on currently running PlexTrac application
+        check                                checks for version & status of PlexTrac application
+        configure                            does initial configuration required for PlexTrac application
+        info                                 display information about the current PlexTrac Instance
+
+### For a fresh install of a new PlexTrac instance:
+
+    DOCKER_HUB_KEY=${DOCKER_HUB_KEY} plextrac configure
+    plextrac update
+    plextrac start
+
+## Testing inside a Vagrant box
+
+This repo is bundled with a [Vagrant](https://www.vagrantup.com/) configuration so that the CLI can be tested inside a virtual machine without polluting your host machine.
+
+You'll need to have [Vagrant](https://www.vagrantup.com/) installed before you can use the preconfigured virtual machine.
+
+1.  Start up the Vagrant virtual machine:
+
+        vagrant up
+
+2.  SSH into the Vagrant VM and become the `plextrac` user
+
+        vagrant ssh
+        sudo -iu plextrac
+
+3.  Create a default `.env` file by running the `configure` command
+
+        plextrac configure
+
+4.  Set your DOCKER_HUB_KEY environment variable in the `.env` file
+
+        vim /opt/plextrac/.env
+
+    You should update the following vars:
+
+        DOCKER_HUB_KEY    # Your token for pulling images from the plextrac private Docker Hub registry
+        UPGRADE_STRATEGY  # A docker image tag to pull (i.e. 'stable', 'edge', 'imminent', or a git commitish)
+
+    You should also set any feature flag or `OVERRIDE` vars in the `.env` file now, _for example_:
+
+        OVERRIDE_TRANSACTION_ID_LOGGING=true
+
+5.  Start the Plextrac application
+
+        plextrac start
+
+6.  You can tail logs:
+
+    ...for a specific service:
+
+        docker-compose logs -f <SERVICE_NAME>
+
+    Example:
+
+        docker-compose logs -f plextracapi
+
+    ...or for the entire stack all at once:
+
+        docker-compose logs -f
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ This is the PlexTrac management CLI that can be used to perform the initial setu
     DOCKER_HUB_KEY=${DOCKER_HUB_KEY} plextrac install -y
 
 ### A quick note about the `DOCKER_HUB_KEY`
+
 It should look several groups of alphanumeric characters separated by hyphens. It should _not_ be base64-encoded.
 Example: `abcd123-11a1-22bb-c3c3-defg567890`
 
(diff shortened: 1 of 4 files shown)


static/docker-compose.yml
```diff
@@ -91,18 +127,8 @@ volumes:
...
-      type: 'none'
-      o: 'bind'
-      device: '${PLEXTRAC_HOME:-.}/volumes/redis'
-  couchbase-backups:
-    driver: local
-    driver_opts:
-      type: 'none'
-      o: 'bind'
-      device: '${PLEXTRAC_HOME:-.}/backups/couchbase'
+  redis: {}
```
evanhsu, line 130: The bind mounts were causing docker-compose to fail while creating the volumes inside the Vagrant box (inside my MacOS host machine).  We may have hit the inception limit here.

It doesn't look like we're using the named volumes anyway and this change fixes the problem.

Vagrantfile
```diff
@@ -34,7 +34,7 @@ Vagrant.configure("2") do |config|
 
   # Create a private network, which allows host-only access to the machine
   # using a specific IP.
-  config.vm.network "private_network", ip: "192.168.13.37"
+  config.vm.network "private_network", ip: "192.168.56.37"
```
evanhsu, line 37: These IP changes are described in their own PR: #2

Vagrantfile
```diff
@@ -72,7 +72,7 @@ Vagrant.configure("2") do |config|
 
   echo ""
   echo "# Example customized deployment directory and domain name:"
-  echo "#   PLEXTRAC_HOME=/var/apps/plextrac-demo CLIENT_DOMAIN_NAME=192.168.13.37 ./plextrac initialize"
+  echo "#   PLEXTRAC_HOME=/var/apps/plextrac-demo CLIENT_DOMAIN_NAME=192.168.56.37 ./plextrac initialize"
```
evanhsu, line 75: These IP changes are described in their own PR: #2

README.md
```diff
@@ -1 +1,72 @@
...
+
+    You should update the following vars:
+
+        DOCKER_HUB_KEY    # Your token for pulling images from the plextrac private Docker Hub registry
+        UPGRADE_STRATEGY  # A docker image tag to pull (i.e. 'stable', 'edge', 'imminent', or a git commitish)
+
+    You should also set any feature flag or `OVERRIDE` vars in the `.env` file now, _for example_:
+
+        OVERRIDE_TRANSACTION_ID_LOGGING=true
+
```
evanhsu, line 55: Secret step #4b:
  - I was copying the `docker-compose.yml` file directly into the `/opt/plextrac` folder inside the Vagrant machine because I couldn't figure out how to use the `DOCKER_COMPOSE_ENCODED` variable

README.md
```diff
@@ -1 +1,72 @@
...
+## Testing inside a Vagrant box
+
+This repo is bundled with a [Vagrant](https://www.vagrantup.com/) configuration so that the CLI can be tested inside a virtual machine without polluting your host machine.
+
+You'll need to have [Vagrant](https://www.vagrantup.com/) installed before you can use the preconfigured virtual machine.
+
+1.  Start up the Vagrant virtual machine:
+
+        vagrant up
+
```
evanhsu, line 33: Secret step #1b:
  - I was copying the `docker-compose.yml` file directly into the `/opt/plextrac` folder inside the Vagrant machine because I couldn't figure out how to use the `DOCKER_COMPOSE_ENCODED` variable

README.md
```diff
@@ -1 +1,72 @@
 # plextrac-manager-util
+
+This is the PlexTrac management CLI that can be used to perform the initial setup of a new PlexTrac installation as well as performing maintenance tasks on running instances.
+
+## Usage
+
+    Usage: plextrac COMMAND [FLAGS]
+
+    [+] Available commands:
```
evanhsu, line 9: I pulled this list directly from the `help` output, but it looks like a few commands were/are missing (`logs`?).

static/docker-compose.yml
```diff
@@ -80,6 +80,42 @@ services:
...
+    image: "plextrac/plextracapi:${UPGRADE_STRATEGY:-stable}"
+    depends_on:
+      - plextracdb
+      - redis
+    restart: always
+    environment:
+      CB_API_PASS: "${CB_API_PASS:?err}"
+      CB_API_USER: "${CB_API_USER:?err}"
+      COUCHBASE_URL: "http://plextracdb"
+      CLIENT_DOMAIN_NAME: ${CLIENT_DOMAIN_NAME:-err}
```
BadgerOps, line 109: 👍

static/docker-compose.yml
BadgerOps: Is this going to break anything on the production Linux hosts? 🤔  cc @jesseops

static/docker-compose.yml
```diff
@@ -80,6 +80,42 @@ services:
...
+      - plextracdb
+      - redis
+    restart: always
+    environment:
+      CB_API_PASS: "${CB_API_PASS:?err}"
+      CB_API_USER: "${CB_API_USER:?err}"
+      COUCHBASE_URL: "http://plextracdb"
+      LOG_LEVEL: "${LOG_LEVEL:-error}"
+      REDIS_CONNECTION_STRING: "${REDIS_CONNECTION_STRING:-redis}"
+      REDIS_PASSWORD: "${REDIS_PASSWORD:?err}"
```
rkeeler, line 95: @evanhsu I think we need CLIENT_DOMAIN_NAME to be required for the Engine here too because the Engine is responsible for constructing the content for the emails, and therefore will need the domain name to create the urls.

rkeeler: Disregard. I got confused. Just double checked the code and it is only the notification-sender service that actually needs this env var.

static/docker-compose.yml
evanhsu: Here's the corresponding change in the dev docker-compose: https://github.com/PlexTrac/product-core-backend/pull/575/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R75

static/docker-compose.yml
jesseops: Yes, we definitely don't want to drop these as we'll move into undefined territory. @evanhsu can you post error logs? Within vagrant this _shouldn't_ behave any differently between host machines since it's running in a VM either way.

evanhsu: I removed the changes to the named volumes from this PR and I'll post some error logs shortly once my vagrant box rebuilds

evanhsu: In an unsurprising turn of events, rebooting my laptop twice appears to have resolved this issue 🤦🏻

static/docker-compose.yml
```diff
@@ -91,7 +127,7 @@ volumes:
   dbdata: {}
   uploads: {}
   letsencrypt: {}
-  redis:
+  redis:
```
BadgerOps, line 130: 👀

jesseops: :thonk:

README.md
```diff
@@ -1 +1,78 @@
...
+        backup                               perform backup on currently running PlexTrac application
+        check                                checks for version & status of PlexTrac application
+        configure                            does initial configuration required for PlexTrac application
+        info                                 display information about the current PlexTrac Instance
+
+## For a fresh install of a new PlexTrac instance:
+
+    DOCKER_HUB_KEY=${DOCKER_HUB_KEY} plextrac configure
+    plextrac update
+    plextrac start
```
jesseops, line 22:
```suggestion
    DOCKER_HUB_KEY=${DOCKER_HUB_KEY} plextrac install -y
```

jesseops:
We figured out the issue with couchbase impacting full testing - I didn't move the docs from the deprecated repo to this one, and my "shortcut" in the vagrant provisioning script didn't call the command I assumed it did... the one that sets the apiuser credentials in couchbase ;)

static/docker-compose.yml
evanhsu: this is how I pad my github stats.

evanhsu: in my defense, this friggin yaml file doesn't adhere to any known convention so I can't use autoformatting

jesseops:
One thing left that I can think of, and I need to run through testing first - how to control the rollout of the new services to avoid getting ahead of the release. I'll :thonk: on that a bit tomorrow (main concern is rolling out something like a health test that isn't in all releases yet). If I were to roll this today we'd immediately break any new setups.

static/docker-compose.yml
```diff
@@ -80,6 +80,50 @@ services:
...
+    environment:
+      CB_API_PASS: "${CB_API_PASS:?err}"
+      CB_API_USER: "${CB_API_USER:?err}"
+      COUCHBASE_URL: "http://plextracdb"
+      CLIENT_DOMAIN_NAME: ${CLIENT_DOMAIN_NAME:?err}
+      LOG_LEVEL: ${LOG_LEVEL:-error}
+      REDIS_CONNECTION_STRING: "${REDIS_CONNECTION_STRING:-redis}"
+      REDIS_PASSWORD: "${REDIS_PASSWORD:?err}"
+      NOTIFICATION_DRY_RUN: ${NOTIFICATION_DRY_RUN}
+      serviceConfig: ${serviceConfig}
```
jesseops, line 120: Are these already available? I _think_ notification_dry_run may need added but I'm not familiar with serviceConfig.

evanhsu:
Next Steps:

We will wait to merge this until the code is available in the `stable` and `edge` images (otherwise the healthcheck command will fail because the `npm run healthcheck:notification-engine` script doesn't exist yet.

static/docker-compose.yml
evanhsu: The `NOTIFICATION_DRY_RUN` env var does not need to be present - an 'undefined' value is treated as a `false` in code.

README.md
jesseops: This is automatically handled when performing the `plextrac configure` command 👍🏼

jesseops (approved):

src/_info.sh
```diff
@@ -30,6 +30,11 @@ function mod_info() {
   info "Disk Statistics"
   msg `check_disk_capacity`
   msg `info_backupDiskUsage`
+  echo >&2 ""
+
+  title "Notifications"
+  info "Summary"
+  msg "`compose_client exec plextracapi npm run status:notifications | fgrep -v '> '`"
```
jesseops, line 37: This drops a few extra newlines in the info output. Try chomping the leading un-used lines (`awk` is helpful here).

jesseops (approved):

PR #4, opened 2022-01-07
Adds logging for events (eg, config updates), various minor cleanup/tweaks

jesseops:


jesseops:
@evanhsu more fysa than anything

PR #19, opened 2022-03-16
SUDO-265 Adds RHEL/CentOS Support & Pre-Checks

jesseops:


src/_check.sh
```diff
@@ -0,0 +1,109 @@
...
+    title "Running checks on installation at '${PLEXTRAC_HOME}'"
+    requires_user_plextrac
+    info "Checking Docker Compose Config"
+    compose_client config -q && info "Config check passed"
+    pending=`composeConfigNeedsUpdated || true`
+    if [ "$pending" != "" ]; then
+        error "Pending Changes:"
+        msg "    %s\n" "$pending"
+    fi
+    VALIDATION_ONLY=1 configure_couchbase_users
```
jesseops, line 21: existing logic that I wrote, but I don't like it.

This _should_ be a standalone function that checks for invalid/unconfigured users. Then we can either print it (as we do here) or go back and "fix" them (as the function currently behaves if the validation only flag is not set)

src/_check.sh
```diff
@@ -0,0 +1,109 @@
...
+  fi
+}
+
+###
+
+# Check for an existing installation
+function _check_no_existing_installation() {
```
jesseops, line 28: This could be updated to use `_precheck_...` prefix, enabling the use of inspection to run all prechecks instead of manually calling them.

src/_cli_common_utilities.sh
```diff
@@ -42,7 +42,7 @@ function event__log_activity() {
   local activity_name="${1:-func:${FUNCNAME[1]}}"
   local activity_data="${2:--}"
 
-  debug `printf "Logged event '%s' at %s\n" $activity_name $activity_timestamp | tee -a "${event_log_filepath}"`
+  debug "`printf "Logged event '%s' at %s\n" $activity_name $activity_timestamp | tee -pa "${event_log_filepath}" 2>&1 || echo "Unable to write to event log"`"
```
jesseops: todo: port change from other tee usage here

jesseops (commented):
comments from virtual code review session

PR #23, opened 2022-06-15
Just use script name instead of hardcoding the expected path

jesseops:
Also, update docs to be in sync with common usage

### DIFF
--- a/src/_initialize_user.sh
+++ b/src/_initialize_user.sh
@@ -26,7 +26,7 @@ function copy_scripts() {
   info "Copying plextrac CLI to user PATH..."
   tmp=`mktemp -p /tmp plextrac-XXX`
   debug "tmp script location: $tmp"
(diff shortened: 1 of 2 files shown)


src/_initialize_user.sh
```diff
@@ -26,7 +26,7 @@ function copy_scripts() {
   info "Copying plextrac CLI to user PATH..."
   tmp=`mktemp -p /tmp plextrac-XXX`
   debug "tmp script location: $tmp"
-  debug "`$(dirname $0)/plextrac dist 2>/dev/null > $tmp && cp -uv $tmp "${PLEXTRAC_HOME}/.local/bin/plextrac"`"
+  debug "`$0 dist 2>/dev/null > $tmp && cp -uv $tmp "${PLEXTRAC_HOME}/.local/bin/plextrac"`"
```
jesseops, line 29: Tested working in Vagrant

PR #24, opened 2022-07-26
CLOUDOPS-10 stub out postgres support stuff

jesseops:
This should be a start if @rkeeler wants to take a stab at it.

I already have all of this working in the UAT environment, but I manually added the secrets (using the script provided here) to the .env. I _think_ the way I have the volumes setup is at least close to optimal, but I've slept once or twice since doing this build.

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -31,6 +31,7 @@ services:
     environment:
       CB_API_PASS: ${CB_API_PASS}
       CB_API_USER: ${CB_API_USER}
+      OVERRIDE_RUNBOOKS_V2: true
       REDIS_CONNECTION_STRING: ${REDIS_CONNECTION_STRING:-redis}
       REDIS_PASSWORD: ${REDIS_PASSWORD:?err}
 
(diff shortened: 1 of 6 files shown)


static/docker-compose.yml
```diff
@@ -148,10 +149,25 @@ services:
...
+    - .env
+    environment:
+      POSTGRES_HOST_AUTH_METHOD: scram-sha-256
+      POSTGRES_INITDB_ARGS: '--auth-local=scram-sha-256 --auth-host=scram-sha-256'
+      PGDATA: /var/lib/postgresql/data/pgdata
+    volumes:
+    # this is where the initdb script & SQL template go
+    - /opt/plextrac/volumes/postgres-initdb:/docker-entrypoint-initdb.d
+    - postgres-data:/var/lib/postgresql/data
+
```
rkeeler, line 165: 👍  These all match the values we've been testing in k8s.

evanhsu:
@jesseops @rkeeler I see that this PR is intended to support the RUNBOOKS_V2 alpha, but is this also intended to support the Dual-Write rollout?  If so, I can review it to help move forward - are there any outstanding todo's that need to be addressed before this can come out of draft status?

jesseops:
> @jesseops @rkeeler I see that this PR is intended to support the RUNBOOKS_V2 alpha, but is this also intended to support the Dual-Write rollout? If so, I can review it to help move forward - are there any outstanding todo's that need to be addressed before this can come out of draft status?

I pulled this all from what I had manually put in place for the runbooks uat environment as a reference, was out most of last week. Tomorrow is the earliest I'll get to moving this PR forward, it needs some work.

jesseops:
These configs are working in vagrant. More needs to happen, but this should get postgres deployed and allow for further testing.

PR #28, opened 2022-08-08
QUICK-576 wire up pg schema migrations and etl

rkeeler:
The postgres db container was recently rolled out, but our application code is not referencing it. This PR wires things up so that the postges schema migrations and ETL are triggered as part of the migration job. And the tenant dual write/read env var is wired up so that we can do the canary rollout by changing those env vars in the .env file and not having to ship new infra changes.

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -2,6 +2,10 @@ version: "3.8"
 
 services:
   plextracapi:
+    depends_on:
+    - plextracdb
+    - redis
+    - postgres
     environment:
       STARTUP_MODE: API_ONLY
     env_file:
@@ -26,6 +30,7 @@ services:
     depends_on:
     - plextracdb
     - redis
+    - postgres
     image: "plextrac/plextracapi:${UPGRADE_STRATEGY:-stable}"
     volumes:
     - uploads:/usr/src/plextrac-api/uploads:rw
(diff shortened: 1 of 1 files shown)


static/docker-compose.yml
```diff
@@ -27,17 +27,32 @@ services:
     - plextracdb
     - redis
     image: "plextrac/plextracapi:${UPGRADE_STRATEGY:-stable}"
+    volumes:
+    - uploads:/usr/src/plextrac-api/uploads:rw
```
rkeeler, line 31: This is needed because we are writing the ETL full output json to the `uploads/etl-logs` folder so we can find detailed info if we need to troubleshoot anything during the rollout. This can go away after that.

static/docker-compose.yml
```diff
@@ -27,17 +27,32 @@ services:
...
     entrypoint: ""
     command: |
       sh -c
         "npm run maintenance:enable &&
          npm run db:migrate &&
+         npm run pg:migrate &&
+         npm run pg:etl up all &&
          npm run maintenance:disable"
     environment:
+      COUCHBASE_URL: "http://plextracdb"
```
rkeeler, line 41: I guess the `db:migrate` script was applying a default value for this. But our ETL script is not.

static/docker-compose.yml
```diff
@@ -27,17 +27,32 @@ services:
     - plextracdb
     - redis
```
jesseops, line 28: Might need to include a dependency link on postgres here.

jesseops:
Just sanity checking - for stable, if the feature flag for dual write is not enabled the etl will just silently pass, right? If so I'm comfortable cutting a full release right now.

jesseops (approved):

PR #29, opened 2022-08-16
This is a mega hack, but nginx does not update dns automatically and …

jesseops:
…short-term this is the simplest way to address issues when the frontend is not updated along with the backend


JordanJStone (approved):

mmorgan-pt (approved):
Verified branch with local testing. LGTM

jesseops:
This was a hack in case I couldn't figure out how to stop Nginx from permanently caching DNS. PRs are in review to handle this and should be going out soon.

PR #39, opened 2022-10-18
QUICK-692 run pg schema migrations before couchbase migrations

rkeeler:
### _What_
Changes the order of the api startup flow so that postgres schema migrations run _before_ couchbase migrations.

### _Why_
The runbooks v2 feature is built to rely on the _couchbase_ migration runner to seed data into _postgres_. While this isn't ideal, we are going to support this for now because we don't have another alternative in place for the Oaks team to use to do this data seeding at the moment. So in order for the couchbase migration runner to be able to seed data into the runbooks postgres tables, the tables will need to exist. Therefore, we now need the pg schema migrations to run before the couchbase migrations.

## Testing & Repro Steps
1. Deploy this to an environment
2. Watch the order of operations (pg schema migrations first, then cb migrations, then etl)

## Jira Issues
Please put links to the Jira issues that prompted this PR
- https://plextrac.atlassian.net/browse/QUICK-692

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -42,8 +42,8 @@ services:
     command: |
       sh -c
         "npm run maintenance:enable &&
-         npm run db:migrate &&
          npm run pg:migrate &&
+         npm run db:migrate &&
          npm run pg:etl up all &&
          npm run maintenance:disable"
     environment:


jesseops:
Plan to give this a test at some point, not likely today though. I'm happy to 👍🏼 if one of y'all has gone through setting up a fresh vagrant instance though!

rkeeler:
@jesseops I haven't tested it in Vagrant yet. I think i'll be able to get to that tomorrow morning. I figured we'd save this one for last and get the core-backend and k8s changes in first.

evanhsu (approved):
I reviewed the logs from the vagrant runs and this looks like it does what we intended.
I approve, but we can wait for @jesseops to give a 👍🏻 before merging.

jesseops (approved):

jesseops:
I'll bump this and release early next week

PR #40, opened 2022-10-25
Mdb/add validate postgres metrics user

Michael-Burke:
- Added function to create `postgres_exporter` user in PG database; this allows scraping to be done at minimum privilege
- Exposed postgres ports to scraping (observability change)

### DIFF
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -19,6 +19,8 @@ function mod_check() {
         msg "    %s\n" "$pending"
     fi
     VALIDATION_ONLY=1 configure_couchbase_users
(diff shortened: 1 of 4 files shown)


jesseops:
I got 
> /opt/plextrac/.local/bin/plextrac: line 1147: PG_METRICS_USER: unbound variable

so I think the check on line 95 needs updated. I tried a few methods but the easiest was to just default to an empty string and check equality 🤷🏼‍♂️

jesseops (approved):
This works and created the desired user only when enabled via presence of envvar.

PR #42, opened 2022-11-03
QUICK-722 Remove default envs vars from couchbase-migration container

evanhsu:
# This PR is in draft status until we confirm that using empty strings for these env vars will give us the intended behavior in the app code

## What
Changes the default values to be _**empty strings**_ for the `PG_TENANTS_WRITE_MODE` and `PG_TENANTS_READ_MODE` env vars in the `couchbase-migration` container.  ~It's now possible for these env vars to be "undefined" in the docker environment.~

## Why
1. The default values for these env vars is being managed in the application code because there's some logic that needs to be applied before choosing a default.
2. The outage described by https://plextrac.atlassian.net/browse/QUICK-721 resulted from the wrong default being applied for these env vars. Removing the defaults from the docker config will help to consolidate all the logic into one spot in the application code.

## Testing
1. Make sure that you don't have either of these env vars defined in your `.env` file.
2. Create a PlexTrac instance using the PlexTrac manager utility
3. Verify that the values being used for these env vars match the defaults set in `libs/dbDualWrite/src/dualWriteConfig.ts`:
    a. You should see logs indicating that the "Tenants ETL" was SKIPPED (because the write_mode defaulted to couchbase_only)
    b. Change the default WRITE_MODE (in the `dualWriteConfig.ts`) to `dual_write` and restart the 'couchbase-migration' container.  You should see that the "Tenants ETL" runs this time because the write_mode defaulted to "dual_write".


jesseops:
As-is this leaves big angry output
image

The fix is to specify an empty default, pretty similar to `bash` envvar handling: `${MYENVVAR-}` will not error and will still pass the override if it's provided. See [`CB_BUCKET`](https://github.com/PlexTrac/plextrac-manager-util/blob/release/v0.3.x-bash/static/docker-compose.yml#L76) for an example.

static/docker-compose.yml
```diff
@@ -60,8 +60,8 @@ services:
       PG_RUNBOOKS_ADMIN_PASSWORD: ${PG_RUNBOOKS_ADMIN_PASSWORD:?err}
       PG_RUNBOOKS_ADMIN_USER: ${PG_RUNBOOKS_ADMIN_USER:?err}
       PG_RUNBOOKS_DB: ${PG_RUNBOOKS_DB:?err}
-      PG_TENANTS_WRITE_MODE: ${PG_TENANTS_WRITE_MODE:-couchbase_only}
-      PG_TENANTS_READ_MODE: ${PG_TENANTS_READ_MODE:-couchbase_only}
+      PG_TENANTS_WRITE_MODE: ${PG_TENANTS_WRITE_MODE}
+      PG_TENANTS_READ_MODE: ${PG_TENANTS_READ_MODE}
```
jesseops, line 64:
```suggestion
      PG_TENANTS_WRITE_MODE: ${PG_TENANTS_WRITE_MODE:-}
      PG_TENANTS_READ_MODE: ${PG_TENANTS_READ_MODE:-}
```

jesseops (changes requested):
see previous comment

static/docker-compose.yml
evanhsu: I see - this makes sense since env vars are always strings - there's no way to define an env var as "undefined"

jesseops:
@runia1 @evanhsu saw this got closed; wondering what's the status of envvars for ETL stuff? Wanna make sure everything is running as everyone expects.

evanhsu:
dear god how old is this?
thanks for the ping Jesse - I don't believe we need this anymore

PR #43, opened 2022-11-09
DATA-62 Adds basic datalake maintainer service + volume config to support rollout to Cloud Hosted instances

jesseops:
Adds the initial `datalake-maintainer` service in a disabled state. This must be scaled up (see https://github.com/PlexTrac/infra-saltstack/pull/99) to actually start the service.

Changes
- Adds a volume to stored the credentials required by the API and datalake services
- Adds the new service in a disabled state

Testing
- Validated in vagrant that the service doesn't start by default
- Added an override and validated that the service starts up

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -18,6 +18,7 @@ services:
     restart: always
     volumes:
     - uploads:/usr/src/plextrac-api/uploads:rw
+    - datalake-maintainer-keys:/usr/src/plextrac-api/keys/gcp:r
     healthcheck:
       test:
       # Handle cURL being removed due to upstream vuln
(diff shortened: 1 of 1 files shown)


static/docker-compose.yml
```diff
@@ -169,6 +170,41 @@ services:
...
+      COUCHBASE_URL: "http://plextracdb"
+      GCP_DATALAKE_SA_KEY_PATH: "${GCP_DATALAKE_SA_KEY_PATH:-/usr/src/plextrac-api/keys/gcp/gcp-datalake-sa-key.json}"
+      LOG_LEVEL: ${LOG_LEVEL:-info}
+      REDIS_CONNECTION_STRING: "${REDIS_CONNECTION_STRING:-redis}"
+      REDIS_PASSWORD: "${REDIS_PASSWORD:?err}"
+      PG_CORE_DB: ${PG_CORE_DB:?err}
+      PG_CORE_RO_PASSWORD: ${PG_CORE_RO_PASSWORD:?err}
+      PG_CORE_RO_USER: ${PG_CORE_RO_USER:?err}
+      PG_CORE_RW_PASSWORD: ${PG_CORE_RW_PASSWORD:?err}
+      PG_CORE_RW_USER: ${PG_CORE_RW_USER:?err}
```
jesseops, line 195: Hey @evanhsu in chatting with Ben he's only doing reads against Postgres (via the service layer). Has the `RO` user been hooked up to service those requests or is it still using `RW`?

evanhsu: As of right now, only the RW user is used to establish a connection to Postgres:
https://github.com/PlexTrac/product-core-backend/blob/main/src/db/core/index.ts#L17

[I'm tracking this conversation in Slack](https://plextrac.slack.com/archives/C02BSEY93T5/p1668024770169389)
and we have this Jira story that could be pulled in if necessary: https://plextrac.atlassian.net/browse/QUICK-675

jesseops: For now I'll leave this one as-is (eg, the `RW` user will be provided) but that can be rolled back when the `RO` user is pushed in. AFAIK the `datalake-maintainer` is just accessing service methods so updating those to use the `RO` account will do the trick.

This is definitely something that must happen soon before we take on more tech debt with the `RW` account, but not an emergency.

PR #44, opened 2022-11-11
QUICK-726

rkeeler:
I threw this PR up to help with the conversation happening in this [slack thread](https://plextrac.slack.com/archives/C022F318D3R/p1668183958316389) so that everyone understands the nature of the issue. 

This is just for discussion for now. We need to validate with logs from the affected environment still, and perhaps test this fix in that environment with a docker-compose override file.


static/docker-compose.yml
```diff
@@ -60,6 +60,8 @@ services:
       PG_CORE_DB: ${PG_CORE_DB:?err}
       PG_RUNBOOKS_ADMIN_PASSWORD: ${PG_RUNBOOKS_ADMIN_PASSWORD:?err}
       PG_RUNBOOKS_ADMIN_USER: ${PG_RUNBOOKS_ADMIN_USER:?err}
+      PG_RUNBOOKS_RW_PASSWORD: ${PG_RUNBOOKS_RW_PASSWORD:?err}
+      PG_RUNBOOKS_RW_USER: ${PG_RUNBOOKS_ADMIN_USER:?err}
```
ruchiraagrawal, line 64: is the password correct?

rkeeler: I think so. This is wiring up the environment variable so that it will be received by the docker container. The actual password will be set in an environment variable on each host environment.

evanhsu:
```suggestion
      PG_RUNBOOKS_RW_USER: ${PG_RUNBOOKS_RW_USER:?err}
```

rkeeler: Ahh I see. Yes good catch!

rkeeler:
Closing because this was fixed and merged separately [here](https://github.com/PlexTrac/plextrac-manager-util/pull/45)

jesseops:
oh thanks @rkeeler I missed that you had prepped this. Had a few threads going on the subject 😬

PR #49, opened 2022-12-06
increase timeouts, silently kill progress bar and retry containers

mmorgan-pt:
**Background**
Small tweaks to handle new instances since they take a long time to run migrations.
Also adds a retry for failed containers to help auto recover on updates.
**Impact**
No errors thrown if the progress bar exits before migrations container does.
Attempts to start containers again if any of them are in a failed state `exited (1)|unhealthy|created` and then throws an error if the issue persists.
**Risk**
An unknown bug could prevent updates or correct installs
**Testing**
1. Tested in a vagrant environment with fresh installs. 
Progress bar finishes running, but the program continues to wait for migration container to finish. Kill task attempts to kill progress bar, but silently fails because the PID no longer exists. Result now is  long progress bar, followed by `done`

### DIFF
--- a/src/_docker_manager.sh
+++ b/src/_docker_manager.sh
@@ -44,3 +44,22 @@ function docker_createInitialComposeOverrideFile() {
   fi
   log "Done"
 }
+
+function checkFailedContainers() {
+  # check current status of containers and restart any that show exited (1), created or unhealthy
+  # if that's found, run the update again
+  if [ `compose_client ps | egrep 'exited \(1\)|unhealthy|created'` ]; then
+    info "An error occured with one or more containers, attempting to start again"
+    # sleep for 5 to give things a moment to settle, then try starting again
+    sleep 5
+    mod_start
+    # check again, then throw an error if containers are still in a bad state
+    if [ `compose_client ps | egrep 'exited \(1\)|unhealthy|created'` ]; then
+      error "One or more containers are in a failed state, please contact support!"
+      return 1
+    fi
+  else
+    debug "no failed containers found, continuing"
+    return 0
+  fi
+}
--- a/src/_update.sh
+++ b/src/_update.sh
@@ -23,6 +23,7 @@ function mod_update() {
   pull_docker_images
   mod_start
   mod_check
+  checkFailedContainers
 }
 
 function _selfupdate_refreshReleaseInfo() {
--- a/src/plextrac
+++ b/src/plextrac
@@ -277,13 +277,13 @@ function mod_start() {
   # it should take an optional param which is a function
   # that should return 0 when ready
   (
-    for i in `seq 1 30`; do
+    for i in `seq 1 60`; do
       progressBar=`printf ".%.0s%s"  {1..$i} "${progressBar:-}"`
       msg "\r%b" "${GREEN}[+]${RESET} ${NOCURSOR}${progressBar}"
       sleep 2
     done &
-    timeout --preserve-status 60 docker wait "$(compose_client ps couchbase-migrations -q)" >/dev/null || { error "Migrations exceeded timeout"; exitStatus=1; }
-    kill $!; trap 'kill $!' SIGTERM
+    timeout --preserve-status 300 docker wait "$(compose_client ps couchbase-migrations -q)" >/dev/null || { error "Migrations exceeded timeout"; exitStatus=1; }
+    ps $! >/dev/null 2>&1 && kill $! >/dev/null 2>&1; trap 'kill $!' SIGTERM
     >&2 echo -n "${RESET}"
     if [ "${exitStatus:-0}" -ne 0 ]; then
       return "$exitStatus"


src/_update.sh
```diff
@@ -23,6 +23,7 @@ function mod_update() {
   pull_docker_images
   mod_start
   mod_check
+  checkFailedContainers
```
jesseops, line 26: Example of grouping the logic here:

```
maxTries=2
for attempt in {1..$maxTries}; do
  mod_start

  unhealthyContainers=$(checkFailedContainers 2>&1)
  if [[ $? -eq 0 ]]; then
    debug "No failed containers found, continuing"
    break # exit the loop if all is kosher
  fi
  debug "${unhealthyContainers}"

  if [[ $attempt >= $maxTries ]]; then
    error "Max attempts failed, some containers are still unhealthy"
    exit 1

  else;
    info "An error occured with one or more containers, attempting to start again"
    sleep 5
  fi
mod_check
```

jesseops: of course, you can just sub in the `compose_client ps` here and no extra function.

jesseops (commented):
Left some thoughts on here. Overall looks good, I'll give it a test later when I'm past some GCP stuff.

src/_docker_manager.sh
```diff
@@ -44,3 +44,22 @@ function docker_createInitialComposeOverrideFile() {
...
+  # if that's found, run the update again
+  if [[ `compose_client ps | egrep 'exited \(1\)|unhealthy|created'` ]]; then
+    info "An error occured with one or more containers, attempting to start again"
+    # sleep for 5 to give things a moment to settle, then try starting again
+    sleep 5
+    mod_start
+    # check again, then throw an error if containers are still in a bad state
+    if [[ `compose_client ps | egrep 'exited \(1\)|unhealthy|created'` ]]; then
+      error "One or more containers are in a failed state, please contact support!"
+      return 1
```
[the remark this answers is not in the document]
mmorgan-pt: we certainly can't have a smelly script, so I refactored and put it in a for loop within the update function rather than having it's own.

src/_docker_manager.sh
```diff
@@ -6,7 +6,7 @@ postgresComposeService="postgres"
 
 function compose_client() {
   flags=($@)
-  compose_files=$(for i in `ls ${PLEXTRAC_HOME}/docker-compose*.yml`; do printf " -f %s" "$i"; done )
+  compose_files=$(for i in `ls -r ${PLEXTRAC_HOME}/docker-compose*.yml`; do printf " -f %s" "$i"; done )
```
jesseops, line 9: I added this as an incidental - the order matters, and these were being applied in literally the _reverse_ I intended 😂 didn't cause an issue that I noticed until I tried to override the image for plextracdb and it didn't take.

src/_update.sh
```diff
@@ -21,7 +21,25 @@ function mod_update() {
...
+
+  # Sometimes containers won't start correctly at first, but will upon a retry
+  maxRetries=2
+  for i in $( seq 1 $maxRetries ); do
+    mod_start
+
+    # Check for failed containers and continue in loop if any are found, otherwise break out of loop
+    compose_client ps | egrep 'exited \(1\)|unhealthy|created' >/dev/null || break
+
+    if [[ $i -ge $maxRetries ]]; then
```
jesseops, line 33: Moved the check above the other log/sleep so it doesn't log "gonna retry" and then dump. Also, swapped to using `>=` (`-ge`) as it potentially saves us from a bug if this code changes down the road.

src/plextrac
```diff
@@ -251,7 +251,7 @@ function mod_install() {
   log "Done"
   mod_autofix
   pull_docker_images
-  mod_start
+  mod_start 360 # allow up to 6 minutes for startup on install, due to migrations
```
jesseops, line 254: 6 minutes _I think_ is about right, but maybe bump to 10 just to be safe?

src/plextrac
```diff
@@ -271,20 +271,31 @@ function mod_start() {
   # Enable database migrations on startup
   compose_client --profile=database-migrations up -d --remove-orphans
 
-  info "Waiting for application startup"
+  waitTimeout=${1:-60}
+  info "Waiting up to ${waitTimeout}s for application startup"
```
jesseops, line 275: Set expectations for the user

src/plextrac
```diff
@@ -271,20 +271,31 @@ function mod_start() {
...
   local progressBar
   # todo: extract this to function waitForCondition
   # it should take an optional param which is a function
   # that should return 0 when ready
   (
-    for i in `seq 1 30`; do
+    while true; do
```
jesseops, line 281: I really hate having this be unbounded, BUT so far in my testing it has not managed to avoid getting killed.

jesseops (approved):

PR #50, opened 2023-01-11
QS-279 Pull the PG_MIGRATE_TENANT env var into the couchbase-migrations service definition

evanhsu:
WHAT:
Makes the new env var `PG_MIGRATE_TENANT` available to the `couchbase-migrations` service container.
(The `plextracapi` service also uses this env var, but that service pulls the entire .env file into the environment so we don't need to explicitly update docker-compose file to mention it).

WHY:
Quicksilver is building the new fault-tolerant ETL-runner, which will no longer require 2 env vars per domain (i.e. PG_TENANT_READ_MODE and PG_TENANT_WRITE_MODE). This PR puts the new, single env var into place that will replace them (PG_MIGRATE_TENANT)

There's [another PR here](https://github.com/PlexTrac/infra-saltstack/pull/128) in the SaltStack repo that will ensure that all cloud-hosted environments have this env var in their `.env` file.

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -65,6 +65,7 @@ services:
       PG_RUNBOOKS_DB: ${PG_RUNBOOKS_DB:?err}
       PG_TENANTS_WRITE_MODE: ${PG_TENANTS_WRITE_MODE:-couchbase_only}
       PG_TENANTS_READ_MODE: ${PG_TENANTS_READ_MODE:-couchbase_only}
+      PG_MIGRATE_TENANT: ${PG_MIGRATE_TENANT} # Default value defined in app code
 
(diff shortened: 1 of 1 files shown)


evanhsu:
Chatted with @jesseops about this and we decided that it's not necessary to make this anymore since the compose overrides have been added to SaltStack: https://github.com/PlexTrac/infra-saltstack/pull/128/files#diff-cceab761d97e4eb48241433c5bd5dc8513d2b2c15ffb582d7e55948960c0fc8dR31

We can manage our own cloud-hosted instances with that method, and if we absolutely need to have an on-prem customer toggle an env var, they can do it in their `docker-compose.override.yml` file

PR #52, opened 2023-01-19
Added Maintenance Mode Check

Michael-Burke:
- Modified the _check.sh and _info.sh files to include a new function to curl the `${CLIENT_DOMAIN_NAME}/api/v2/health/full` endpoint and return current maintenance mode status

### DIFF
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -20,6 +20,7 @@ function mod_check() {
     fi
     VALIDATION_ONLY=1 configure_couchbase_users
     postgres_metrics_validation
+    check_for_maintenance_mode
 
 
     # echo >&2 ""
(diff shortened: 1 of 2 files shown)


src/_check.sh
```diff
@@ -29,6 +30,18 @@ function mod_check() {
     # msg "`compose_client exec plextracapi npm run status:notifications | fgrep -v '> '`"
   fi
 }
+function check_for_maintenance_mode() {
+  title "Checking Maintenance Mode"
+  IN_MAINTENANCE="null"
+  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode)
```
jesseops, line 36:
```suggestion
  IN_MAINTENANCE=$(curl -Lks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode)
```

jesseops: probably not _critical_ but this was helpful testing locally

jesseops:
`inMaintenanceMode` is not present during install :(

src/_check.sh
jesseops: Ok it all works out now that the install is complete

jesseops (approved):
This looks good

PR #53, opened 2023-01-27
Set Postgres port to bind only on localhost

jesseops:
## Background

We opened up this port so we could get metrics, but missed in review that this binds to _all_ ports.

## Impact

Makes the Postgres port only bind on _localhost_

## Risk

Will cause a Postgres restart when this applies

## Testing

Run an update with this version and observe postgres is bound only to localhost

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -243,7 +243,7 @@ services:
     - postgres-data:/var/lib/postgresql/data
     - postgres-backups:/backups
     ports:
-    - 5432:5432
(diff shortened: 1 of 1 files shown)


jesseops:
Screenshot from 2023-01-27 10-35-44

image

image

Looks like this doesn't impact the app, as expected (the app hits the container directly)

Michael-Burke (approved):
LGTM; simple change restricting the port binding

BadgerOps (approved):
good catch here

m4l1c3 (approved):
💪

PR #54, opened 2023-01-27
QS-308 pg restore retain ownership

rkeeler:
## Overview
Fixes a bug where taking a postgres backup and then restoring that backup will cause future pg schema migrations to fail because the ownership of the tables will be wrong. Our schema migration runner expects that the individual db admin users will be owners of the tables, but after performing a pg restore the pg super user is the owner instead.

This PR fixes this by removing the CLI flag that was causing the table ownership to be ignored when performing the pg_restore.

## Steps to reproduce
1. Take a backup: `plextrac backup`
2. Restore that backup: `plextrac restore`
3. Shell into the pg container: `docker-compose exec postgres /bin/bash`
4. Open a psql connection: `PGPASSWORD=$PG_CORE_ADMIN_PASSWORD psql -U $PG_CORE_ADMIN_USER core`
5. Check the table owner: `\dt`

**Buggy behavior**
```
                 List of relations
 Schema |        Name        | Type  |    Owner     
--------+--------------------+-------+--------------
 public | tenant             | table | internalonly
```

**Correct behavior**
```
                List of relations
 Schema |        Name        | Type  |   Owner    
--------+--------------------+-------+------------
 public | tenant             | table | core_admin
```

### DIFF
--- a/src/_restore.sh
+++ b/src/_restore.sh
@@ -76,8 +76,10 @@ function restore_doPostgresRestore() {
       log "Extracting backup for $db"
       debug "`compose_client exec -T $postgresComposeService\
         tar -xvzf /backups/$backupFile ./$db.psql 2>&1`"
-      log "Restoring $db"
-      dbRestoreFlags="-d $db --clean --if-exists --disable-triggers --verbose"
+      dbAdminEnvvar="PG_${db^^}_ADMIN_USER"
+      dbAdminRole=$(eval echo "\$$dbAdminEnvvar")
+      log "Restoring $db with role:${dbAdminRole}"
+      dbRestoreFlags="-d $db --clean --if-exists --no-privileges --no-owner --role=$dbAdminRole  --disable-triggers --verbose"
       debug "`compose_client exec -T -e PGPASSWORD=$POSTGRES_PASSWORD $postgresComposeService \
         pg_restore -U $POSTGRES_USER $dbRestoreFlags ./$db.psql 2>&1`"
       debug "`compose_client exec -T $postgresComposeService \
--- a/src/_restore.sh
+++ b/src/_restore.sh
@@ -77,7 +77,7 @@ function restore_doPostgresRestore() {
       debug "`compose_client exec -T $postgresComposeService\
         tar -xvzf /backups/$backupFile ./$db.psql 2>&1`"
       log "Restoring $db"
-      dbRestoreFlags="-d $db --clean --if-exists --no-owner --disable-triggers --verbose"
+      dbRestoreFlags="-d $db --clean --if-exists --disable-triggers --verbose"
       debug "`compose_client exec -T -e PGPASSWORD=$POSTGRES_PASSWORD $postgresComposeService \
         pg_restore -U $POSTGRES_USER $dbRestoreFlags ./$db.psql 2>&1`"
       debug "`compose_client exec -T $postgresComposeService \


src/_restore.sh
```diff
@@ -77,7 +77,7 @@ function restore_doPostgresRestore() {
       debug "`compose_client exec -T $postgresComposeService\
         tar -xvzf /backups/$backupFile ./$db.psql 2>&1`"
       log "Restoring $db"
-      dbRestoreFlags="-d $db --clean --if-exists --no-owner --disable-triggers --verbose"
```
rkeeler, line 80: @jesseops Do you recall if there was a specific reason that this `--no-owner` flag was being used?

jesseops: My goal here was to handle situations where the database user differed - eg, in some environment `core_admin` was `svc-plextrac-postgres-core-admin` or something similar. The idea being to use the _current_ db setup as the source of truth for roles/permissions. It's possible that this is an unnecessary precaution or one we can resolve through a runbook instead.

jesseops: I was initially thinking that this could be addressed through some kind of role mapping, eg

```
dbAdminRole=<do something to map `core` to $PG_CORE_ADMIN here>
dbRestoreFlags="-d $db --clean --if-exists --role=$dbAdminRole --no-owner --disable-triggers --verbose"
```

rkeeler: I saw your suggestion about `--role=core_admin` in the Jira ticket and that was the first thing i tried. I couldn't get it to work and not sure exactly what i was doing wrong.

jesseops: Here's what I came up with (I ran into a snag where I was using `DB_<db>_ADMIN` instead of `PG_<db>_ADMIN_USER` :doh:)

Screenshot from 2023-01-27 14-43-09
Screenshot from 2023-01-27 14-42-53

jesseops: :embarrassed: ok maybe `--role` and `--no-owner` are incompatible.
Screenshot from 2023-01-27 14-46-12

jesseops: OH well if you do a restore with one user and THEN try to set role to another it doesn't happy. :badpokerface: trying a new test case where

1. Postgres is freshly setup
2. Backup taken as normal
3. Verify table owner is `core_admin`
4. Perform restore using `internalonly` user but drop role to `core_admin`
5. Validate table owner remains the same

jesseops: I _think_ that works

image

jesseops: https://github.com/PlexTrac/plextrac-manager-util/pull/55/files happy to chat through this path and pros/cons

jesseops: UGH well turns out it's more complex than all that because the permissions are included in the backup :/ 
image

jesseops: Ok, if I add `--no-privileges` the restore finishes and I see the correct table owner. Investigating if the privileges are correct all the way down due to the original permission grants or if this is an issue

image

jesseops: I _think_ that does it 
image

rkeeler: Nice! Pulling in your changes to test them now

jesseops: This was definitely the area I was most concerned about when putting together the initial restore stuff, so it's a) a bummer it did end up having issues and b) satisfying to _hopefully_ have the desired end result functioning, more or less 😂 


Let me know how it goes. If this is still an issue we can just move to the change you had already and have a runbook for changing the owner/grants if needed.

rkeeler: Sweet, this worked perfectly. I was able to perform the `plextrac restore` and have the table ownership be correct. And i validated that by performing a schema migration ('down' and then back 'up') and the core_admin and runbooks_admin users were able to modify the schema without issue.

This looks ready to go from my testing. @jesseops Feel free to approve and merge once you are comfortable with it on your end.

src/_restore.sh
```diff
@@ -76,8 +76,10 @@ function restore_doPostgresRestore() {
       log "Extracting backup for $db"
       debug "`compose_client exec -T $postgresComposeService\
         tar -xvzf /backups/$backupFile ./$db.psql 2>&1`"
-      log "Restoring $db"
-      dbRestoreFlags="-d $db --clean --if-exists --no-owner --disable-triggers --verbose"
+      dbAdminEnvvar="PG_${db^^}_ADMIN_USER"
```
rkeeler, line 79: Ahh yes, `^^`. Completely intuitive way to convert to upper case 🤣 . Oh bash, never change.

jesseops (approved):
Collab'ed and worked out a good solution that should be reasonably flexible and avoid the ownership issue in the future. :shipit:

jesseops:
Thanks @rkeeler!

PR #58, opened 2023-01-31
Fix for failing `plextrac install` command

Michael-Burke:
- Added logic on `plextrac install` command to limit user validation inside DBs until after they are started
- Moved Maintenance mode check in `plextrac info` command to make more sense

### DIFF
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -33,15 +33,9 @@ function mod_check() {
 }
 function check_for_maintenance_mode() {
   title "Checking Maintenance Mode"
-  IN_MAINTENANCE="null"
-  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode) || IN_MAINTENANCE="null"
-  if [ $IN_MAINTENANCE == "true" ]; then
-    info "Maintenance Mode: $IN_MAINTENANCE"
-  elif [ $IN_MAINTENANCE == "false" ]; then
-    info "Maintenance Mode: $IN_MAINTENANCE"
-  else
-    info "Maintenance Mode: Error"
-  fi
+  IN_MAINTENANCE="Unknown"
+  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode) || IN_MAINTENANCE="Unknown"
+  info "Maintenance Mode: $IN_MAINTENANCE"
 }
 
 ###
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -18,11 +18,9 @@ function mod_check() {
         error "Pending Changes:"
         msg "    %s\n" "$pending"
     fi
-    if [ ${INSTALL_ONLY:-0} -eq 0 ]; then
-      VALIDATION_ONLY=1 configure_couchbase_users
-      postgres_metrics_validation
-      check_for_maintenance_mode
-    fi
+    VALIDATION_ONLY=1 configure_couchbase_users
+    postgres_metrics_validation
+    check_for_maintenance_mode
 
     # echo >&2 ""
 
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -31,7 +31,6 @@ function mod_check() {
 }
 function check_for_maintenance_mode() {
   title "Checking Maintenance Mode"
-  IN_MAINTENANCE="Unknown"
   IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode) || IN_MAINTENANCE="Unknown"
   info "Maintenance Mode: $IN_MAINTENANCE"
 }
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -18,10 +18,11 @@ function mod_check() {
         error "Pending Changes:"
         msg "    %s\n" "$pending"
     fi
-    VALIDATION_ONLY=1 configure_couchbase_users
-    postgres_metrics_validation
-    check_for_maintenance_mode
-
+    if [ ${INSTALL_ONLY:-0} -eq 0 ]; then
+      VALIDATION_ONLY=1 configure_couchbase_users
+      postgres_metrics_validation
+      check_for_maintenance_mode
+    fi
 
     # echo >&2 ""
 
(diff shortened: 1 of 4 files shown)


src/_check.sh
```diff
@@ -18,10 +18,11 @@ function mod_check() {
         error "Pending Changes:"
         msg "    %s\n" "$pending"
     fi
-    VALIDATION_ONLY=1 configure_couchbase_users
-    postgres_metrics_validation
-    check_for_maintenance_mode
-
+    if [ ${INSTALL_ONLY:-0} -eq 0 ]; then
+      VALIDATION_ONLY=1 configure_couchbase_users
```
jesseops, line 22: What about moving this logic up a layer? eg, just put the user creation stuff into the install function. Then drop the `mod_check` call from `mod_install`

jesseops (changes requested):
Logic concerns

src/_check.sh
```diff
@@ -33,7 +34,7 @@ function mod_check() {
 function check_for_maintenance_mode() {
   title "Checking Maintenance Mode"
   IN_MAINTENANCE="null"
-  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode)
+  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode) || IN_MAINTENANCE="null"
```
jesseops, line 37: I think that below could be simplified to `info "Maintenance Mode: $IN_MAINTENANCE"` + 
```suggestion
  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode) || IN_MAINTENANCE="unknown"
```

jesseops: fewer lines of code and less logic to thing about

src/_info.sh
```diff
@@ -26,13 +26,14 @@ function mod_info() {
   msg "    %s\n" "$active"
   echo >&2 ""
 
+  #Check for Maintenance Mode
+  msg `check_for_maintenance_mode`
```
jesseops, line 30: I missed this before but - I don't think you need to wrap this in the `msg` plus backticks, as the function itself is already calling `info`.

src/plextrac
```diff
@@ -237,7 +237,7 @@ function mod_install() {
   title "Installing PlexTrac Instance"
   requires_user_plextrac
   mod_configure
-  mod_check
+  INSTALL_ONLY=1 mod_check
```
jesseops, line 240: I think I would just do the direct call to validate the docker syntax here instead of running the whole `mod_check` since `mod_check` by design assumes a running system.

src/_check.sh
jesseops: edit: I got confused, I mean _don't_ wrap this stuff in logic to determine if an install is running or not, just have `mod_install` in the _pre-install_ step do docker compose validation instead of calling the `mod_check` function. That way it is doing less work and we don't have branching logic.

src/plextrac
```diff
@@ -302,10 +302,9 @@ function mod_start() {
 
 function mod_autofix() {
   title "Fixing Auto-Correctable Issues"
-  configure_couchbase_users
   # Add postgres configuration monitor here
   postgres_metrics_validation
-  
+  configure_couchbase_users
```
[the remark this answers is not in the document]
Michael-Burke: There was some confusing on this function not returning to complete. Reverting changes

src/plextrac
Michael-Burke: Copied logic from mod_check into the install to validate compose file before DBs are spun up.

src/_check.sh
Michael-Burke: The preinstall flag is never used IIRC. When the instance is stood up, `plextrac --preinstall` is never called and in the normal workflow it isn't called either. In fact its not used outside of this context:
```bash
# only used for mod_check
      "--pre" | "--preinstall")
        DO_PREINSTALL_CHECKS=1
        shift
        ;;
```

This also means that these two functions don't get run:

```bash
if [ ${DO_PREINSTALL_CHECKS:-0} -eq 1 ]; then
    title "Running pre-installation checks"
    _check_system_meets_minimum_requirements
    _check_no_existing_installation
  else
    title "Running checks on installation at '${PLEXTRAC_HOME}'"
```

Regardless of that, I've copied the compose file validation to the `mod_install` function so it will validate that before starting the DBs

jesseops: Yep, the `preinstall` checks are for "readiness" of a host system and would be manually run by the end user fysa

src/_check.sh
```diff
@@ -32,15 +31,9 @@ function mod_check() {
...
-  if [ $IN_MAINTENANCE == "true" ]; then
-    info "Maintenance Mode: $IN_MAINTENANCE"
-  elif [ $IN_MAINTENANCE == "false" ]; then
-    info "Maintenance Mode: $IN_MAINTENANCE"
-  else
-    info "Maintenance Mode: Error"
-  fi
+  IN_MAINTENANCE="Unknown"
+  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode) || IN_MAINTENANCE="Unknown"
+  info "Maintenance Mode: $IN_MAINTENANCE"
```
jesseops, line 36:
```suggestion
  IN_MAINTENANCE=$(curl -ks https://${CLIENT_DOMAIN_NAME}/api/v2/health/full | jq .data.inMaintenanceMode || "Unknown")
  info "Maintenance Mode: $IN_MAINTENANCE"
```

does this match as an equivalent?

src/_info.sh
```diff
@@ -26,13 +26,14 @@ function mod_info() {
   msg "    %s\n" "$active"
   echo >&2 ""
 
+  #Check for Maintenance Mode
+  msg check_for_maintenance_mode
```
jesseops, line 30: the logging cmd should happen in only one place, so I'd say drop the `msg` here and let the `info` in the function itself handle things.
```suggestion
  check_for_maintenance_mode
```

src/plextrac
```diff
@@ -237,7 +237,13 @@ function mod_install() {
...
   requires_user_plextrac
   mod_configure
-  mod_check
+  info "Checking Docker Compose Config"
+  compose_client config -q && info "Config check passed"
+  pending=`composeConfigNeedsUpdated || true`
+  if [ "$pending" != "" ]; then
+      error "Pending Changes:"
+      msg "    %s\n" "$pending"
+  fi
```
jesseops, line 246:
```suggestion
```

this logic can be omitted since all we want to check here is that the Compose syntax is valid

jesseops:
Threw in some comments/suggestions @Michael-Burke - happy to huddle up if you'd like to go through it. I again apologize for the dirty bash foo you're navigating.

jesseops:
Screenshot from 2023-02-01 11-50-13
Screenshot from 2023-02-01 11-50-02

jesseops (approved):
Looks good, overall this is cleaner and provides more helpful feedback to users!

PR #59, opened 2023-02-02
[CLOUDOPS-191] Update default `.env` value

Michael-Burke:
- Changed default value for .env key DOCKER_HUB_KEY to `ptcustomers` per request CLOUDOPS-191


jesseops (approved):
This is fine - we manage these values for cloud instances so this is just needed for self-hosted installs. If we ever change the user then we'll need to also update this value.

PR #64, opened 2023-03-03
Use a temporary volume for uploads backup

jesseops:
## Background

This is some old logic that is super inefficient and was triggering some false alarms in monitoring. Simple change to avoid some extra filtering we have to do and make our security team less twitchy.

## Impact

Minor change to the mechanics of taking a backup with no change to outputs

## Risk

N/a

## Testing

Run a backup `/vagrant/src/plextrac backup -v` and validate the correct command is used and backup structure is present.


StevenPaugh (approved):
LGTM

PR #65, opened 2023-03-08
Adds nginx healthcheck and corrects startup check to use case insensitive match

jesseops:
## Background

We occasionally see updates "complete" but Docker Compose has only _created_ the container. This change ensures that scenario is detected, and further enables health checks to ensure the nginx proxy has started up and is able to negotiate SSL connections

## Impact

Will cause the nginx proxy container to be recreated and adds health check

## Risk

Low, the health check works with self signed certificates and largely exists just to ensure the startup script has completed.

## Testing

Tested in vagrant with self signed cert and evaluated scenarios:
- Invalid container config (causing a restart loop)
- Unable to retrieve LetsEncrypt cert (unable to setup SSL connection)
- Connection failure (specified bad port)

### DIFF
--- a/src/_update.sh
+++ b/src/_update.sh
@@ -28,14 +28,14 @@ function mod_update() {
     mod_start
 
     # Check for failed containers and continue in loop if any are found, otherwise break out of loop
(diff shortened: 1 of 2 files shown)


Michael-Burke (approved):
LGTM

PR #66, opened 2023-04-04
Add error handling for mod_start

jesseops:
This should address the situations where docker-compose fails to start containers that have been created, leading to manual recovery steps

### DIFF
--- a/src/_update.sh
+++ b/src/_update.sh
@@ -25,7 +25,7 @@ function mod_update() {
   # Sometimes containers won't start correctly at first, but will upon a retry
   maxRetries=2
   for i in $( seq 1 $maxRetries ); do
-    mod_start
+    mod_start || true
 
     # Check for failed containers and continue in loop if any are found, otherwise break out of loop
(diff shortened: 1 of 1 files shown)


src/_update.sh
```diff
@@ -25,7 +25,7 @@ function mod_update() {
   # Sometimes containers won't start correctly at first, but will upon a retry
   maxRetries=2
   for i in $( seq 1 $maxRetries ); do
-    mod_start
+    mod_start || true
```
jesseops, line 28: Took me awhile to realize that _Docker Compose_ is throwing an error on `up` which is why the below checks aren't getting triggered in the particular case they're meant to handle. This is a mega simple change to keep from going into cleanup immediately, and instead retry the update.

jesseops:
```suggestion
    mod_start || sleep 5 # Wait before going on to health checks, they should handle triggering retries if mod_start errors
```

jesseops: @mmorgan-pt thoughts?

PR #67, opened 2023-04-26
Adds support for restoring from backups during the install

jesseops:
## Background

Data restoration is important and stuff. Also it sucks to be slow.

## Impact

Allows smoother restores by handling the data restore during the install step. This reduces extra manual commands that have to be run, and handles large datasets more easily as the indexes do not yet exist.

## Risk

Low, the capability is behind a flag that is not documented so this is additive and not disruptive to current workflows.

## Testing

I tested in a local vagrant environment, with a backup I had available (couchbase only, I constructed a fake uploads backup as well). I verified that the correct files are restored, the appropriate containers exist _prior_ to running the restore, postgres/uploads are optional.

### DIFF
--- a/src/_restore.sh
+++ b/src/_restore.sh
@@ -32,6 +32,9 @@ function restore_doUploadsRestore() {
 
 function restore_doCouchbaseRestore() {
   title "Restoring Couchbase from backup"
+  debug "Fixing permissions"
+  debug "`compose_client exec -T $couchbaseComposeService \
+    chown -R 1337:1337 /backups 2>&1`"
   latestBackup="`ls -dc1 ${PLEXTRAC_BACKUP_PATH}/couchbase/* | head -n1`"
   backupFile=`basename $latestBackup`
   info "Latest backup: $latestBackup"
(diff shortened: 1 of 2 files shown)


src/_restore.sh
```diff
@@ -47,7 +50,7 @@ function restore_doCouchbaseRestore() {
 
     log "Running database restore"
     compose_client exec $couchbaseComposeService cbrestore /backups http://localhost:8091 \
-      -u ${CB_BACKUP_USER} -p "${CB_BACKUP_PASS}" -x conflict_resolve=0,data_only=1
+      -u ${CB_BACKUP_USER} -p "${CB_BACKUP_PASS}" --from-date 2022-01-01 -x conflict_resolve=0,data_only=1
```
jesseops, line 53: This should allow us to skip the "nuke the old 2020 backup" step.

src/_restore.sh
```diff
@@ -32,6 +32,9 @@ function restore_doUploadsRestore() {
 
 function restore_doCouchbaseRestore() {
   title "Restoring Couchbase from backup"
+  debug "Fixing permissions"
+  debug "`compose_client exec -T $couchbaseComposeService \
```
jesseops, line 36: Necessary to allow us to extract the backup.

Michael-Burke (approved):
LGTM! :shipit:

PR #68, opened 2023-04-26
Adds a workaround to cbrestore blowing chunks when running a restore non-interactively

jesseops:
Tested using local vagrant, as well as "live" on a dev box using a manually edited script. Both normal and non-interactive modes work (eg, SSH and saltstack invocation)

```shell
vagrant ssh -c "sudo -i -u plextrac /vagrant/src/plextrac restore -y -v" -- -T
```

Screenshot from 2023-04-26 17-41-24

### DIFF
--- a/src/_restore.sh
+++ b/src/_restore.sh
@@ -49,7 +49,9 @@ function restore_doCouchbaseRestore() {
       tar -xzvf /backups/$backupFile 2>&1`"
 
     log "Running database restore"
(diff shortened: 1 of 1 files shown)


Michael-Burke:
Did we test that its still functional with a simple `plextrac restore` without flags?

Michael-Burke (approved):
LGTM

PR #69, opened 2023-04-28
Fixes for RHEL installations

mbaileyPT:
These changes make it so that customers using Red Hat will not have to alter the utility at all.


jesseops (changes requested):
Hey @mbaileyPT IIRC you were going to schedule time to review this, correct? I do not have a RHEL system handy to test myself, and these changes appear to drop support for CentOS (rocky linux now?).

mbaileyPT:
In it's current state, the utility doesn't work for CentOS or RHEL, so these are focused solely on getting RHEL working through the utility. I planned on taking a look at the CentOS side later, but with the influx of RHEL installation we've been getting, it makes sense to have at least RHEL working as opposed to neither working.

jesseops (approved):
This has been hanging awhile and doesn't affect our supported installation mode, so I'm fine with the change.

PR #72, opened 2023-06-01
PG-99: Add env var

nicholasvrdoljak:
## What
This adds the API_INTEGRATION_AUTH_CONFIG_NOTIFICATION_SERVICE env var to the config generator. It also gives the notification engine access to the .env file so that it can access this variable. 

## Why
This will allow inter-service network requests and is necessary to decouple the notification engine from couchbase in preparation for postgres.

### DIFF
--- a/src/_configure_plextrac.sh
+++ b/src/_configure_plextrac.sh
@@ -28,6 +28,7 @@ function generate_default_config() {
 
   # Generate base env, using imported vars from above where applicable
   generatedEnv="
+API_INTEGRATION_AUTH_CONFIG_NOTIFICATION_SERVICE=${API_INTEGRATION_AUTH_CONFIG_NOTIFICATION_SERVICE:-`generateSecret`}
 JWT_KEY=${JWT_KEY:-`generateSecret`}
 MFA_KEY=${MFA_KEY:-`generateSecret`}
 COOKIE_KEY=${COOKIE_KEY:-`generateSecret`}
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -135,9 +135,7 @@ services:
       LOG_LEVEL: "${LOG_LEVEL:-info}"
       REDIS_CONNECTION_STRING: "${REDIS_CONNECTION_STRING:-redis}"
       REDIS_PASSWORD: "${REDIS_PASSWORD:?err}"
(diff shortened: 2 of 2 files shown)


static/docker-compose.yml
```diff
@@ -135,6 +135,9 @@ services:
       LOG_LEVEL: "${LOG_LEVEL:-info}"
       REDIS_CONNECTION_STRING: "${REDIS_CONNECTION_STRING:-redis}"
       REDIS_PASSWORD: "${REDIS_PASSWORD:?err}"
+    # Specifically needs the API_INTEGRATION_AUTH_CONFIG_NOTIFICATION_SERVICE generated in src/_configure_plextrac.sh
+    env_file:
+    - .env
```
runia1, line 140: @evanhsu I can't remember how we wrote the engine/sender but we should verify it's only the engine that needs to make api calls.

src/_configure_plextrac.sh
```diff
@@ -28,6 +28,7 @@ function generate_default_config() {
 
   # Generate base env, using imported vars from above where applicable
   generatedEnv="
+API_INTEGRATION_AUTH_CONFIG_NOTIFICATION_SERVICE=${API_INTEGRATION_AUTH_CONFIG_NOTIFICATION_SERVICE:-`generateSecret`}
```
jesseops, line 31: Will this be a separate key for each service? I don't know how the app uses it, but I want to be sure this works the way you expect.

static/docker-compose.yml
jesseops, line 140: We actually don't want to include the `.env` @nicholasvrdoljak -- just the required envvars. Otherwise we get envvar creep and have to unwind which envvars are needed by which service. Simply add the required envvar to the `environment` section above.

jesseops (changes requested):

src/_configure_plextrac.sh
nicholasvrdoljak: Yes, a separate key for each service that needs to use one should be generated.

static/docker-compose.yml
nicholasvrdoljak: Done! See latest commit.

jesseops:
image
Looks like the envvar is generated properly and made available to the notification-engine service.

jesseops (approved):

PR #74, opened 2023-06-22
SSL Certificate Renewal Command

Michael-Burke:
I got tired of not having a way to handle this with our utility so I expanded the functionality

- Added script to handle certificate reload of SSL Certificates into NGINX
- This should handle various circumstances of .env state with the values in various states or missing

**Note:**
I'm open to changing the name of the command from `plextrac reload-cert` to something else; let me know your suggestions.


src/_cert_renew.sh
```diff
@@ -0,0 +1,35 @@
...
+  if [ "$LETS_ENCRYPT_EMAIL" != '' ] && [ "$USE_CUSTOM_CERT" == 'false' ]; then
+    # IF LETS_ENCRYPT = TRUE
+    # ASK TO REMOVE pem/key
+    info "Let's Encrypt certificate detected!"
+    info "Would you like to force renew your SSL certificate?"
+    if get_user_approval; then
+      #if yes, continue to delete pem/key and recreate NGINX
+      info "Deleting current certificates and loading new certificates"
+      compose_client rm -f plextracnginx
+      docker volume rm plextrac_letsencrypt
```
jesseops, line 17: In most cases we can simply renew without needing to nuke our local config. This is a bit heavy-handed; if you want to force renewal even before the 30d expiration window this would be better handled w/ invoking acme.sh (probably).

jesseops (commented):
I'm more or less ok with adding a wrapper to make cert renewal easier, but I _think_ it should be simpler (just recreate the container, no mucking about w/ the volume needed). Am I missing something that requires nuking the volume?

src/_cert_renew.sh
Michael-Burke: Basically, if you force-recreate the NGINX container, if your certificates are still valid, it will simply remount them and go on its way. Either I need to:
1. Remove the volume entirely (most dramatic)
1. Remove the certificates from the volume so they can be regenerated by NGINX
1. Exec into the container, and run acme.sh with force renew

I wrote this at night time, so I'll rework to use Acme to force-renew the certs from inside the NGINX

jesseops: More specifically - do we have a use case for forcing lets encrypt renewal? If that's not a common use case then we should avoid making that the default behavior here.

Michael-Burke: Removed ability to force renew

jesseops (dismissed):

jesseops (approved):

PR #75, opened 2023-06-22
Allow using current version of docker-compose

jesseops:
This reverts a pin we'd previously placed due to a bug in an earlier version of docker-compose. This will only affect _new_ installations, as the install only runs during `plextrac init`.

I tested our workflows including instance setup (from scratch) and this update looks good to go.

### DIFF
--- a/src/_setup_packages.sh
+++ b/src/_setup_packages.sh
@@ -116,7 +116,7 @@ function install_docker() {
 function install_docker_compose() {
   info "upgrading docker-compose..."
   curl -sL $(curl -sL \
-    https://api.github.com/repos/docker/compose/releases/tags/v2.2.3 | jq -r \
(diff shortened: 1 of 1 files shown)


src/_setup_packages.sh
```diff
@@ -116,7 +116,7 @@ function install_docker() {
 function install_docker_compose() {
   info "upgrading docker-compose..."
   curl -sL $(curl -sL \
-    https://api.github.com/repos/docker/compose/releases/tags/v2.2.3 | jq -r \
+    https://api.github.com/repos/docker/compose/releases/latest | jq -r \
```
jesseops, line 119: I am working on a change to permit pinning via envvar and updating ad-hoc, but it is a larger lift so for now we can ship this and come back around in a later changeset.

PR #82, opened 2023-08-21
ES1-455: Prevent NPM traffic on healthcheck

jfjermestad-plextrac:
# Summary
Our containers make a call to NPM each time the health check runs. Adding this flag to the `npm run` command will prevent the network call to check versions from happening.

# Caveat
This is really hard to test in GitPod, please confirm everything runs smoothly in Edge using the `plextrac-manager-util` before merging this. I've done my best to test usage using alternative methods, but nothing beats testing the actual change.

Screenshot 2023-08-21 at 11 32 34 AM

# Jira
https://plextrac.atlassian.net/browse/ES1-445

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -142,9 +142,12 @@ services:
       - "CMD"
       - "npm"
       - "run"
+      - "--no-update-notifier"
       - "healthcheck:notification-engine"
       - "readiness"
       - "10"
+      - "--"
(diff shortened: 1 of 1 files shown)


mmorgan-pt (approved):
Nice find! I tested a fresh install in a VM test environment and everything comes up correctly and logs show the services come online, so this looks good to me

PR #85, opened 2023-09-29
Update: Permission Check/Fix

Michael-Burke:
- Added permission check/fix for a known permission problem with the `etl-logs` folder

### DIFF
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -37,6 +37,7 @@ function check_for_maintenance_mode() {
 }
 
 function mod_etl_fix() {
+  debug "Running ETL Fix"
   local dir=`compose_client exec plextracapi find -type d -name etl-logs`
   if [ -n "$dir" ]
   then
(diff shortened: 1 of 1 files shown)


Michael-Burke:
@mmorgan-pt fixed the permission issue you described. Will not correctly instruct and check for root / sudo use to fix the issue when present

src/_check.sh
mmorgan-pt, line 1: Line 13 (requires_user_plextrac) collides with the etl_fix for root owned directories. I think moving `mod_etl_fix` above this line would allow it to work. That's just one idea though. When I tested, it prompted me to run as sudo or root, but running as root prompts to run as plextrac user

mmorgan-pt (changes requested):

src/_check.sh
Michael-Burke: Adjusted the need for running as root, since I can exec as UID 0 in the container and adjust permissions there

mmorgan-pt (approved):
Nice find! clever and elegant solution!

PR #86, opened 2023-10-12
Update: Package check / validation

Michael-Burke:
- Check to validate that packages are correct on a `plextrac check`
- Also validation onto `plextrac update` so it will validate needed packages before trying to use them


mmorgan-pt (approved):
Good catch!

PR #88, opened 2023-10-17
Update: `docker compose` vs `docker-compose`

Michael-Burke:
- Reworked to use and check for docker compose correctly

### DIFF
--- a/src/_migrate.sh
+++ b/src/_migrate.sh
@@ -147,14 +147,14 @@ function checkExistingConfigForOverrides() {
   esac
 
   info "Checking legacy configuration"
-  dcCMD="docker-compose -f $legacyComposeFile -f $legacyDatabaseFile"
+  dcCMD="docker compose -f $legacyComposeFile -f $legacyDatabaseFile"
   ${dcCMD} config -q || die "Invalid legacy configuration - please contact support"
 
   decodedComposeFile=$(base64 -d <<<$DOCKER_COMPOSE_ENCODED)
   #diff -N --unified=2 --color=always --label existing --label "updated" $targetComposeFile <(echo "$decodedComposeFile") || return 0
   diff --unified --color=always --show-function-line='^\s\{2\}\w\+' \
     <($dcCMD config --no-interpolate) \
-    <(docker-compose -f - <<< "${decodedComposeFile}" -f $composeOverrideFile config --no-interpolate) || return 0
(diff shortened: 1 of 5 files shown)


mmorgan-pt (approved):
Awesome work! Tested this out in a local environment and it appears to work great!

BadgerOps:
What about on an existing install that has `docker-compose` and won't fall into the `install_docker_compose()` flow?

Michael-Burke:
> What about on an existing install that has `docker-compose` and won't fall into the `install_docker_compose()` flow?

@BadgerOps we're adding in a package validation that will check for `docker compose` prior to updating: https://github.com/PlexTrac/plextrac-manager-util/blob/4cf1aeb14fab22cb5dbcd4ec0b5839b2436c7cff/src/_check.sh#L111

src/_migrate.sh
```diff
@@ -147,14 +147,14 @@ function checkExistingConfigForOverrides() {
...
   ${dcCMD} config -q || die "Invalid legacy configuration - please contact support"
 
   decodedComposeFile=$(base64 -d <<<$DOCKER_COMPOSE_ENCODED)
   #diff -N --unified=2 --color=always --label existing --label "updated" $targetComposeFile <(echo "$decodedComposeFile") || return 0
   diff --unified --color=always --show-function-line='^\s\{2\}\w\+' \
     <($dcCMD config --no-interpolate) \
-    <(docker-compose -f - <<< "${decodedComposeFile}" -f $composeOverrideFile config --no-interpolate) || return 0
+    <(docker compose -f - <<< "${decodedComposeFile}" -f ${composeOverrideFile} config --no-interpolate) || return 0
   return 1
   #diff --color=always -y --left-column <($dcCMD config --format=json | jq -S . -r) <(docker-compose -f - <<< "$decodedComposeFile" -f $composeOverrideFile config --format=json | jq -S . -r) | grep -v '^\+'
```
hwilson-r7, line 159: Simultaneously early and late to the party, having started this work in a branch on my fork a long while back and never getting around to PR'ing it 😬 , and then I missed commenting before this one is merged.

The only difference I see in mine that appears missing in this one is that lines 84, 134 and 159 in `src/_migrate.sh` reference `docker-compose` in log/comments, in case you decide you want to go for consistency.

Michael-Burke: Oh nice catch; will add that into the Bugfix PR

PR #89, opened 2023-10-17
Bugfixes: Various bugfixes

Michael-Burke:
- Adjusted jq value for -r instead of -re
- Adjusted image comparison to work with null tags and use `ID` instead of `RepoDigests` which caused issues if multiple digests
- Adjusted usage to `compose_client` over `docker compose` explicit call

### DIFF
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -108,7 +108,7 @@ function _check_os_supported_flavor_and_release() {
 
 # Check for some base required packages to even validate the system
 function _check_base_required_packages() {
-  requiredCommands=('jq' 'lsb_release' 'wget' 'bc' 'docker')
+  requiredCommands=('jq' 'lsb_release' 'wget' 'bc')
   missingCommands=()
   status=0
   for cmd in ${requiredCommands[@]}; do
--- a/src/_check.sh
+++ b/src/_check.sh
@@ -39,10 +39,10 @@ function check_for_maintenance_mode() {
 
 function mod_etl_fix() {
   debug "Running ETL Fix"
-  local dir=`compose_client exec plextracapi find -type d -name etl-logs`
+  local dir=`docker compose exec plextracapi find -type d -name etl-logs`
   if [ -n "$dir" ]
   then
-    local owner=`compose_client exec plextracapi stat -c '%U' uploads/etl-logs`
+    local owner=`docker compose exec plextracapi stat -c '%U' uploads/etl-logs`
     info "Checking volume permissions"
     if [ "$owner" != "plextrac" ]
       then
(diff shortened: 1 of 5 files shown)


src/_clean.sh
```diff
@@ -66,5 +66,5 @@ function clean_sweepUploadsCache() {
   info "core-backend: Cleaning uploads/exports caches"
   # Leaving the cleanup fairly light, this should help a ton without getting aggressive
   debug "`compose_client exec -T -w /usr/src/plextrac-api/uploads plextracapi \
-    find . -maxdepth 1 -type f -regextype posix-extended -regex '^.*\.(json|xml|ptrac|csv|nessus)$' -delete`"
```
Michael-Burke, line 69: Need to understand better if there was a reason that `-regextype posix-extended` was chosen...

src/_clean.sh
```diff
@@ -66,5 +66,5 @@ function clean_sweepUploadsCache() {
...
+    find . -maxdepth 1 -type f -regex '^.*\.\(json\|xml\|ptrac\|csv\|nessus\)$' -delete`"
```
Michael-Burke, line 69: Gross bash-ism but seems to work

src/_clean.sh
BadgerOps: > Need to understand better if there was a reason that `-regextype posix-extended` was chosen...

https://www.gnu.org/software/findutils/manual/html_node/find_html/posix_002dextended-regular-expression-syntax.html#posix_002dextended-regular-expression-syntax

Michael-Burke: @BadgerOps Thank you!

src/_clean.sh
Michael-Burke: @BadgerOps does this accomplish the same thing in your opinion?

src/_clean.sh
BadgerOps: No problem. Now, why @jesseops added that is still a mystery that doc strings or a semi snarky GitHub comment could possibly fix.

src/_check.sh
```diff
@@ -108,7 +108,7 @@ function _check_os_supported_flavor_and_release() {
 
 # Check for some base required packages to even validate the system
 function _check_base_required_packages() {
-  requiredCommands=('jq' 'lsb_release' 'wget' 'bc')
+  requiredCommands=('jq' 'lsb_release' 'wget' 'bc' 'docker')
```
jesseops, line 111: The utility itself handles docker installation, so ideally shouldn't depend on docker.

Michael-Burke: Totally get the pre-install portion. This isn't complete and we're trying to get it to validate that `docker compose` vs `docker-compose` is working in the environment. That part is incomplete in this PR and a WIP

PR #92, opened 2023-10-25
Adds auto-versioning pipeline to simplify release process

jesseops:
This is a straight copy from https://github.com/jesseops/cicd-auto-versioning -- this allows a release to be automatically created with an appropriate version based on the label applied to the PR.

The labels this workflow looks for are `bump:patch`, `bump:minor`, `bump:major` and the related semver bump is just what you'd expect. If no bump label is applied to the PR, no release is created.

This relies on the GitHub automation user running actions, so this shouldn't require any extra setup. The only extra dependency is the `bump-my-version` python utility.

### DIFF
--- a/.github/workflows/bump-version-on-merge.yaml
+++ b/.github/workflows/bump-version-on-merge.yaml
@@ -1,0 +1,90 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Version Bump
+
+on:
+  pull_request:
+    branches:
+      - "main"
+    types:
+      - "closed"
+
+jobs:
+  version-bump:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    defaults:
+      run:
+        shell: bash
+    timeout-minutes: 2
+    outputs:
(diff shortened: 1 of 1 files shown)


hwilson-r7:
What are your thoughts on adding a step to the workflow which would merge to a `release` branch if the version was bumped?

Use case: I have this plextrac-manager-util repository as a git submodule within my repository, however git submodules only support branches and they sadly do not support tags.

jesseops:
> What are your thoughts on adding a step to the workflow which would merge to a `release` branch if the version was bumped?

@hwilson-r7 should be reasonably simple; just need to update [this line](https://github.com/PlexTrac/plextrac-manager-util/pull/92/files#diff-2494a4ccce3d21b57e9b3f98c256fb6e4f93c9027e8b8b250e650f39a80ba96bR55) to _also_ checkout a new branch and push it.

Something like

```shell
... && git checkout -b release/v`bump-my-version show current_version` && git push -u
```

ought to do the trick.

jesseops:
I just added the ability to manually release a version in addition to the PR auto versioning: https://github.com/jesseops/cicd-auto-versioning/compare/v0.0.3...v0.1.0

jesseops:
😢

PR #93, opened 2023-11-01
Update: Contiguous Updates

Michael-Burke:
## Goal
- Enable contiguous updates to the PlexTrac application starting with `2.0`
  - E.g., if a customer is running `1.55`, the application will update to `2.0`, and then `2.1`, `2.2`, etc to current version
  - E.g., if a customer is running `2.3`, the application will updates to `2.4`, `2.5`, etc to current version
- Ensure that method doesn't block updates for on-prem and air-gapped customers who's app doesn't contact dockerhub at all

## Tasks To-do
- [x] Notate if SKIP_SELF_UPGRADE is checked and handle upgrade with user input and message acknowledging the need for contiguous updates from `2.0` on
- [x] Error handling if JWT Token is empty
- [x] Scope versions fetched for comparison between newest, to running configuration and handle 
- [x] Add logic to check if `2.0` is available to the public
- [x] Add logic to check if the running version is before `2.0` and if so to update straight to `2.0`, and then contiguously after (`1.55 -> 2.0 -> 2.1 -> 2.2 -> etc`)
- [x] Add time estimate and user input to validate that they want to upgrade and take that time
- [x] Updated `Vagrantfile` to use Ubuntu 22.04
- [x] Added hard check to ensure admin's known how long down time will be
- [x] Bump version to 0.5.0

### DIFF
--- a/src/_cli_common_utilities.sh
+++ b/src/_cli_common_utilities.sh
@@ -4,8 +4,9 @@ function get_user_approval() {
   # If non-interactive, log failure and return 1
   # If -y/--assume-yes/ASSUME_YES flags/envvars are set, return 0
   # If -y/--assume-yes/ASSUME_YES flags/envvars are NOT allowed, skip returning 0
+  # If a parameter is passed when the fucntion is called, its assumed this will ignore -y --yes flags
   if [[ -n ${1:-} ]]; then
-      debug "hard-check requested; disallowing return 0 from -y or --assume-yes"
+      debug "hardstop requested; disallowing return 0 from -y or --assume-yes"
     else
       if [ ${ASSUME_YES:-false} == "true" ]; then return 0; fi
   fi
--- a/src/_cli_common_utilities.sh
+++ b/src/_cli_common_utilities.sh
@@ -4,12 +4,7 @@ function get_user_approval() {
   # If non-interactive, log failure and return 1
   # If -y/--assume-yes/ASSUME_YES flags/envvars are set, return 0
   # If -y/--assume-yes/ASSUME_YES flags/envvars are NOT allowed, skip returning 0
-  # If a parameter is passed when the fucntion is called, its assumed this will ignore -y --yes flags
-  if [[ -n ${1:-} ]]; then
-      debug "hardstop requested; disallowing return 0 from -y or --assume-yes"
-    else
-      if [ ${ASSUME_YES:-false} == "true" ]; then return 0; fi
-  fi
+  if [ ${ASSUME_YES:-false} == "true" ]; then return 0; fi
   tty -s || die "Unable to request user approval in non-interactive shell, try passing the -y or --assume-yes CLI flag"
   PS3='Please select an option: '
   select opt in "Yes" "No" "Exit"; do
--- a/src/_cli_common_utilities.sh
+++ b/src/_cli_common_utilities.sh
@@ -3,7 +3,12 @@ function get_user_approval() {
   # If interactive, prompt for user approval & return 0
   # If non-interactive, log failure and return 1
   # If -y/--assume-yes/ASSUME_YES flags/envvars are set, return 0
-  if [ ${ASSUME_YES:-false} == "true" ]; then return 0; fi
+  # If -y/--assume-yes/ASSUME_YES flags/envvars are NOT allowed, skip returning 0
+  if [[ -n ${1:-} ]]; then
+      debug "hard-check requested; disallowing return 0 from -y or --assume-yes"
+    else
+      if [ ${ASSUME_YES:-false} == "true" ]; then return 0; fi
+  fi
   tty -s || die "Unable to request user approval in non-interactive shell, try passing the -y or --assume-yes CLI flag"
   PS3='Please select an option: '
   select opt in "Yes" "No" "Exit"; do
--- a/src/_version_check.sh
+++ b/src/_version_check.sh
@@ -8,7 +8,7 @@ function upgrade_time_estimate() {
   if (( ${#upgrade_path[@]} >= 1 ))
     then
       time_estimate=$(echo "${#upgrade_path[@]} * 15" | bc -l)
-      error "Detected ${#upgrade_path[@]} upgrade(s). Each upgrade can take up to 15 minutes to pull the new version, and update the running application. Given the number of upgrades, the projected upgade time is $time_estimate. Are you sure you want to continue?"
+      error "Detected ${#upgrade_path[@]} upgrade(s). Each upgrade can take up to 15 minutes to pull the new version, and update the running application. Given the number of upgrades, the projected upgade time is $time_estimate minutes. Are you sure you want to continue?"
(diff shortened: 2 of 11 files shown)


src/_version_check.sh
```diff
@@ -0,0 +1,187 @@
...
+            else
+              upstream_tags+=(`wget --header="Authorization: JWT "${JWT_TOKEN} -O - "https://hub.docker.com/v2/repositories/plextrac/plextracapi/tags/?page=$page&page_size=1000" -q | jq -r .results[].name | grep -E '(^[0-9]\.[0-9]*$)' || true`)
+            fi
+            page=$[$page+1]
+        done
+        # Compare stable vs latest_tag and if stable is older than the newest tag, then remove the newest tag to ensure we don't update past stable
+        if [ $stable_date -le $latest_date ]; then
+          debug "Stable is older than latest version"
+          # If stable is older, remove latest from upgrade path
+          for i in ${!upstream_tags[@]}
```
mmorgan-pt, line 127: if upstream_tags is a sorted list, you could just pop the first item rather than iterate through the entire list looking for it. Based on what I'm reading, I _think_ it should be sorted

src/_version_check.sh
```diff
@@ -0,0 +1,187 @@
...
+                  debug "correcting upstream_tags to remove latest"
+                  unset 'upstream_tags[i]'
+              fi
+          done
+        else
+          debug "Stable is pinned to the latest version $latest_ver"
+        fi
+        
+        # Remove the running version from the Upgrade path
+        for i in ${!upstream_tags[@]}
```
mmorgan-pt, line 140: we should be able to make some assumptions here as well if the list is sorted and not have to just iterate through it

src/_version_check.sh
```diff
@@ -0,0 +1,187 @@
...
+          do
+            if [[ ${upstream_tags[i]} = $running_ver ]]
+              then
+                debug "correcting upstream_tags to remove running version"
+                unset 'upstream_tags[i]'
+            fi
+        done
+        for i in "${!upstream_tags[@]}"; do
```
mmorgan-pt, line 148: What wizardry is this? Sorting the list?

mmorgan-pt (changes requested):
Not sure why this is an issue now, but the Volume permissions fix does not appear to be working. I suspect it's caused by a change in the base container, but the user:group `plextrac:plextrac` doesn't exist in the api container, but rather `appuser:appuser` which has the same 1337 uid. 

This is observed on fresh ubuntu 22.04 and 20.04 boxes in vagrant. 

I think running the chown command within the host, rather than in the container, would resolve the issue. Alternatively, setting it to `appuser` or 1337 would probably work too.

src/_version_check.sh
```diff
@@ -0,0 +1,187 @@
...
+    upstream_tags=()
+    page=1
+    if [ "$contiguous_update" = true ]
+      then
+        # Get upstream tag list
+        debug "Looking for Running version $running_ver or Breaking version $breaking_ver"
+        while [ $page -lt 600 ]
+          do
+            # Get the available versions from DockerHub and save to array
+            if [[ $(echo "${upstream_tags[@]}" | grep "$breaking_ver" || true) ]]
```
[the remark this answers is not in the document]
Michael-Burke: It errors if it can't find anything so by always OR evaluating to true it prevents an exit

src/_version_check.sh
```diff
@@ -0,0 +1,187 @@
...
+            elif [[ $(echo "${upstream_tags[@]}" | grep "$running_ver" || true) ]]
+              then
+                debug "Found running version $running_backend_version"; break;
+            else
+              upstream_tags+=(`wget --header="Authorization: JWT "${JWT_TOKEN} -O - "https://hub.docker.com/v2/repositories/plextrac/plextracapi/tags/?page=$page&page_size=1000" -q | jq -r .results[].name | grep -E '(^[0-9]\.[0-9]*$)' || true`)
+            fi
+            page=$[$page+1]
+        done
+        # Compare stable vs latest_tag and if stable is older than the newest tag, then remove the newest tag to ensure we don't update past stable
+        if [ $stable_date -le $latest_date ]; then
```
mmorgan-pt, line 124:
```suggestion
        if [ $stable_date -l $latest_date ]; then
```
looks like having this set to "less than or equal" allowed it to update all the way to 2.0 even though I set upgrade strategy to stable

mmorgan-pt (changes requested):
Noticed this in testing and I _think_ this is why. Going to validate and continue testing

src/_version_check.sh
Michael-Burke: This does in-fact sort the list ⚠️ 🪀

src/_version_check.sh
Michael-Burke: Talked through this live as well; this removes null values and resets the array to contiguous index values instead of inserting null values at the missing indices

src/_version_check.sh
Michael-Burke: This is removed pending other changes to the upstream DockerHUB repo for tighter validation

src/_cli_common_utilities.sh
```diff
@@ -3,7 +3,12 @@ function get_user_approval() {
   # If interactive, prompt for user approval & return 0
   # If non-interactive, log failure and return 1
   # If -y/--assume-yes/ASSUME_YES flags/envvars are set, return 0
-  if [ ${ASSUME_YES:-false} == "true" ]; then return 0; fi
+  # If -y/--assume-yes/ASSUME_YES flags/envvars are NOT allowed, skip returning 0
```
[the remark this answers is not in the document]
Michael-Burke: Cleaned up the language here. Currently reassessing if we need to hardstop at all during the script or that's going to break update automations and the like from customers. ~Benefit vs Gain~

src/_version_check.sh
Michael-Burke: This is a No-Op since we're fixing this in the upstream repo

src/_cli_common_utilities.sh
```diff
@@ -3,7 +3,13 @@ function get_user_approval() {
...
+  # If a parameter is passed when the fucntion is called, its assumed this will ignore -y --yes flags
```
hwilson-r7, line 7:
```suggestion
  # If a parameter is passed when the function is called, its assumed this will ignore -y --yes flags
```
🙃  (I promise I wasn't here just to look for a spelling error and troll you MB... was just looking for something else lol)

Michael-Burke: lol I'm ripping this out anyway because it will break any automation people have setup to run an update 🙃

hwilson-r7: Awww were you thinking of me? NGL that is part of why I was looking at this PR 😅

mmorgan-pt (approved):
Everything looks good. Testing has been solid and all concerns have been addressed. Nice work!

PR #95, opened 2023-11-30
Tweaking src/_rollout.sh to not scale up data lake maintainer during updates

m4l1c3:


mrala (approved):

mmorgan-pt (dismissed):
Tested in a vagrant and confirmed this is working correctly

Michael-Burke (approved):
Added a debug line to help understand if it breaks here. Otherwise, I validated this does successfully skip the datalake maintainer.

PR #97, opened 2023-12-05
QS-756: Standardize envs and entrypoints

runia1:
## Description

The goal of this changeset is to standardize the way environment variables and entrypoints are applied to all the containers which use the backend image. During development these containers share a single .env file and all env vars are present in each process (api, notification-engine, notification-sender, etc,.).  However we do not apply this same pattern in production which can cause confusion and unnecessary complexity when rolling out new services or adding functionality to existing services other than the `plextracapi`.

## Testing Steps

- in an existing docker-compose environment checkout this branch of the manger-util
- Run `plextrac update -v`
- Observe that all services come up and function
- Trigger some action in the UI which will send emails such as resetting your password.
- Verify the email was sent (this will verify both the notification services are working and redis)

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -51,8 +51,6 @@ services:
          npm run maintenance:disable"
     env_file:
     - .env
-    environment:
-      PG_MIGRATE_PATH: /usr/src/plextrac-api
 
   plextracdb:
     environment:
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -28,7 +28,8 @@ services:
       - 'import requests'
       - 'print(requests.get("http://127.0.0.1:4350/api/v2/health/live").json())'
       - 'EOF'
-
+    entrypoint: npm run
+    command: "start"
 
   couchbase-migrations:
     profiles:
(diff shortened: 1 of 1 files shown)


static/docker-compose.yml
```diff
@@ -28,7 +28,8 @@ services:
       - 'import requests'
       - 'print(requests.get("http://127.0.0.1:4350/api/v2/health/live").json())'
       - 'EOF'
-
+    entrypoint: npm run
+    command: "start"
```
runia1, line 32: All other services which use the plextracapi image have their entrypoints defined in this way.  This just makes the entrypoints straightforward for devs trying to make sense of how this container is started.

static/docker-compose.yml
```diff
@@ -48,24 +49,10 @@ services:
...
+    env_file:
+    - .env
     environment:
-      COUCHBASE_URL: ${COUCHBASE_URL:-http://plextracdb}
-      CB_API_PASS: ${CB_API_PASS}
-      CB_API_USER: ${CB_API_USER}
-      REDIS_CONNECTION_STRING: ${REDIS_CONNECTION_STRING:-redis}
-      REDIS_PASSWORD: ${REDIS_PASSWORD:?err}
-      PG_HOST: ${PG_HOST:-postgres}
       PG_MIGRATE_PATH: /usr/src/plextrac-api
```
runia1, line 55: It looks like the code defaults to this path for production builds, so this env is not useful outside of dev environments.

static/docker-compose.yml
```diff
@@ -48,24 +49,8 @@ services:
...
-      PG_CORE_DB: ${PG_CORE_DB:?err}
-      PG_RUNBOOKS_ADMIN_PASSWORD: ${PG_RUNBOOKS_ADMIN_PASSWORD:?err}
-      PG_RUNBOOKS_ADMIN_USER: ${PG_RUNBOOKS_ADMIN_USER:?err}
-      PG_RUNBOOKS_RW_PASSWORD: ${PG_RUNBOOKS_RW_PASSWORD:?err}
-      PG_RUNBOOKS_RW_USER: ${PG_RUNBOOKS_RW_USER:?err}
-      PG_RUNBOOKS_DB: ${PG_RUNBOOKS_DB:?err}
-      PG_TENANTS_WRITE_MODE: ${PG_TENANTS_WRITE_MODE:-couchbase_only}
-      PG_TENANTS_READ_MODE: ${PG_TENANTS_READ_MODE:-couchbase_only}
+    env_file:
+    - .env
```
Michael-Burke, line 53: The reason we don't want to inject all the environment variables into each container is two fold:

1. Security. We want to ensure services only have access to the secrets they need
2. Future Planning. We will eventually break each of the listed services into a discrete service. At that time, we don't want to have an untracked reliance on a secret that could have been prevented by limited the access the service has to the secrets. With this type of information, generally, its better to be explicit than implicit. A scale solution would be using a vault or secrets manager but that's a future state.

This applied to all the other changes of the same kind in this PR.

static/docker-compose.yml
Michael-Burke: I think this is fine but I'll need to test it doesn't conflict with the base container entrypoint for any reason.

PR #99, opened 2023-12-30
Bugfix: Fix for `.profile`, `.bash_profile`, and `.bashrc` Creation Failure

Michael-Burke:
- Rewrote logic around initial `.profile` copying to avoid error's and issues and fix copying the profile files


mmorgan-pt (approved):
Very nice! Tested this and it ran without error and gave expected behavior. One note is while not idempotent, it doesn't need to be given the rest of the script and low risk of copying the skel file each time this is run, regardless of current state of files in the /opt/plextrac directory

PR #103, opened 2024-01-24
Removing definition to avoid causing upgrade issue for customer upgrading to version prior to 2.1

alex-wold:
The Contextual Scoring Service was intended to rollout with version 2.1 and a few canary customers. We made the change to the plextrac-manager-util static `docker-compose.yml` file to include the Contextual Scoring Service definition. Unfortunately, because the plextrac-manager-util is not tied to the specific backend repo version, any customer running the util is getting the same `contextual-scoring-service` definition. This has caused upgrades in environments version `2.0` and lower to fail to complete the update

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -261,37 +261,6 @@ services:
     - 127.0.0.1:5432:5432/tcp
     restart: always
 
-  contextual-scoring-service:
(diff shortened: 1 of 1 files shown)


mmorgan-pt (approved):
Looks good, this will unblock users

PR #104, opened 2024-01-24
Adding the Contextual Scoring Service definition with public image tag

alex-wold:
After looking at the service definition that was added here https://github.com/PlexTrac/plextrac-manager-util/pull/103 it looks like we were referencing an internal docker registry as opposed to the public one that customers should be using for pulling the public `plextracapi` image. 

We've got 36 canary customers running Contextual Scoring with version `2.1` so ideally we'll need to reinstate this service definition before any hotfixes are released to version `2.1` to avoid the Contextual Scoring Service from "disappearing" in their environments via a `plextrac update` command.

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -260,6 +260,37 @@ services:
     ports:
     - 127.0.0.1:5432:5432/tcp
     restart: always
+  
+  contextual-scoring-service:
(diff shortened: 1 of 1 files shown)


Michael-Burke (approved):
Reviewed with Alex and made changes to ensure the Service exits gracefully if the version of code isn't available.

PR #107, opened 2024-01-31
Bugfix/QOL: Rename Containers After Recreation

Michael-Burke:
# Tasks
- [x] Added way for containers to get renamed to prevent N sprawl on naming
- [x] This will also help metrics get tracked correctly by service name na dnot leave disconnected timeseries
- [x] Added handling for container scaling for the contextual scoring service

# Testing
- [x] RHEL
- [x] Ubuntu
- [x] RL
- [x] CentOS


mmorgan-pt (approved):
This could potentially be more robust with an array and nested for loops, but I couldn't quickly get that working, so lets go with this. It works, its readable and addresses the current issue. Nice work!

PR #108, opened 2024-02-01
Bugfix: RHEL 7 Compatibility

Michael-Burke:
- Updated OS Check to validate for RHEL 7 and remove command non-existent in older packages
  -  Specifically around the `diff` command and the `--nobest` flag for `yum`

### DIFF
--- a/src/_cli_common_utilities.sh
+++ b/src/_cli_common_utilities.sh
@@ -81,3 +81,14 @@ function _load_static() {
     export SYSTEM_REQUIREMENTS=`cat "$staticFilesDir/system-requirements.json"`
   fi
 }
+
+function os_check() {
+  OS_NAME=$(grep '^NAME' /etc/os-release | cut -d '=' -f2)
+  OS_VERSION=$(grep '^VERSION_ID' /etc/os-release | cut -d '=' -f2)
+  color_always="--color=always"
+  if echo "$OS_NAME" | grep -q "Red"; then
+    if echo "$OS_VERSION" | grep -q "7"; then
+      color_always=""
+      fi
+  fi
+}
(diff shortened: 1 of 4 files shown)


src/_version_check.sh
```diff
@@ -103,19 +103,18 @@ function version_check() {
         debug "Looking for Running version $running_ver or Breaking version $breaking_ver"
         while [ $page -lt 600 ]
           do
+            upstream_tags+=(`wget --header="Authorization: JWT "${JWT_TOKEN} -O - "https://hub.docker.com/v2/repositories/plextrac/plextracapi/tags/?page=$page&page_size=1000" -q | jq -r .results[].name | grep -E '(^[0-9]\.[0-9]*$)' || true`)
```
Michael-Burke, line 106: This was moved up and should haven't a problem with the validation of the version. Helps populate the version before the if check happens

src/_version_check.sh
```diff
@@ -125,11 +124,16 @@ function version_check() {
...
             new_array+=( "${upstream_tags[i]}" )
         done
-        upstream_tags=("${new_array[@]}")
-        unset new_array
+        if "${#new_array[@]}" > 0; then
+                upstream_tags=("${new_array[@]}")
+                unset new_array
+        else
+                upstream_tags=("")
+        fi
```
Michael-Burke, line 136: This checks if the array length is greater than 0 and then attempts to use it. Otherwise it sets it to an expected value of ""

src/_version_check.sh
```diff
@@ -140,11 +144,14 @@ function version_check() {
...
-        IFS=$' \t\n'
-
+        if "${upstream_tags[@]}" != ""; then
+                # Sort the upstream tags weve chosen as the upgrade path
+                IFS=$'\n' upgrade_path=($(sort -V <<<"${upstream_tags[*]}"))
+                # Reset IFS to default value
+                IFS=$' \t\n'
+        else
+                upgrade_path=("")
+        fi
```
Michael-Burke, line 154: Same thing here; This checks if the array length is greater than 0 and then attempts to use it. Otherwise it sets it to an expected value of ""

mmorgan-pt (approved):
This looks good to me! Nice work!

PR #110, opened 2024-02-16
Remove PG_DEBUG_QUERY_LOGGING env var

wtkersting:
Remove the PG_DEBUG_QUERY_LOGGING environment variable since it is getting incorrectly parsed


Michael-Burke (approved):
This is being removed correctly

PR #111, opened 2024-02-20
Bugfix: Adjusted logic comparison and missed color reference

Michael-Burke:
- Further Support for colorization and better logic comparison


mmorgan-pt (approved):
ahhh nice catch!

PR #116, opened 2024-03-19
Contextual Scoring Service Fixes

StevenPaugh:
PR to fix contextual-scoring service in docker-compose.yml so that the service restarts in all cases of failure (docker service restart, service failure, system reboot, etc..)

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.5.5
+current_version = 0.5.7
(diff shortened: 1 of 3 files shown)


StevenPaugh:
Waiting on OPC team to determine if the following can be removed as well from the docker-compose.yml file:
`      restart_policy:
        condition: always
        delay: 5s
        window: 120s`

mmorgan-pt (approved):
Nice catch! This all looks great!

PR #119, opened 2024-03-21
sp/Added if statement for version removal from docker-compose.override.yml

StevenPaugh:
Adds if statement to sed replace version in the docker-compose.override.yml file if it has not already been overwritten.


mmorgan-pt (approved):
Nice work!

PR #120, opened 2024-03-21
Upgrade: Adding Podman support (0.6.0)

Michael-Burke:
## Description

There are several good reasons to support podman as a container runtime. There have been some security concerns around using Docker in the RHEL distribution and the goal of this upgrade is to support `podman` and eventually, podman-compose.

## Tasks
- [x] Initial buildout to support Podman in RHEL
- [x] Build out install and init functions to support Podman
- [x] Build out support for all command natively with Podman
- [x] Investigate behavior on reboot (resilience)
- [x] QA full suite Testing
- [x] DR full testing                                                               

**Disclaimer:** Tests are were run from blank VMs in Google Cloud ComputeEngine with the provided Cloud Images for the above distributions. Testing was done exclusively on RHEL 9 in rootless mode using Systemd. This is preferred according to RedHat since Podman 5 removes support for CNI networking.

## Notes
The implementation of podman comes with some trade-offs:

- Restarting the host VM will not start the application containers backup. This is due to how podman implements rootless containers and offers systemd unit files as a solution. If this is desired, then user are encouraged to follow this or a similar method to create those systemd unit files and configuration. However, PlexTrac doesn't officially support this and isn't responsible for any side-effects of using this method including data loss. The officially supported method is to run `plextrac start` if the host VM every restarts
- Due to podman using rootless containers and the way in which it does that, using `su` or similar command to switch users followed by interacting with the manager util will break the containers due to how the permissions work. RedHat states that a direct ssh session to the user you want to or are running podman containers as, is the recommended path. There may be work arounds for this including `loginctl enable-linger plextrac` depending on your operating environment.
- RedHat's RHEL 9 uses SELinux enforced which may cause issues depending on the security posture of the host OS. On a base install of RHEL 9 from a cloud provider there is a need to allow binding to port 443 via a `/etc/sysctl.conf` change
  - Add the following to the `/etc/sysctl.conf` file: `net.ipv4.ip_unprivileged_port_start=443`
  - Reload sysctl: `sysctl --system`

## Change Log
- Added environment variable container runtime evaluation on init and configure
- Added functionality for installing `podman` and `podman-plugins` from the native RHEL repos
- Added sub-functionality to `plextrac install` to consider podman
- Added consideration for all runnable commands
- Fixed `plextrac info` to show actual image version instead of the image hash
- Added `plextrac autocomplete` command to enable tab completion of manager utility commands
- Added Redis healthcheck
- Deprecated `migrate` command and removed from help screen

## Tests -- **Passing**
<details>
<b>Podman only supported on RHEL 8/9. Tests were run on the below OS's but podman specifically wasn't tested outside RHEL 8/9</b>

| Complete  | Command | OS Support |
| :-------------: | :------------- | :-------------: |
| ✅   | `autocomplete` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `backup` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `clean` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `dist` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `info` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `install` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `restore` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `stop` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `version` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `autofix` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `check` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `configure` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `help` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `initialize` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `logs` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `reload-cert` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `start` | `[RHEL 9, Ubuntu 22.04]` |
| ✅   | `update` | `[RHEL 9, Ubuntu 22.04]` |

</details>

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.5.5
+current_version = 0.6.0
 commit = True
 tag = True
 
--- a/README.md
+++ b/README.md
@@ -61,17 +61,18 @@ Depending on your configuration, you may need to solve the following issues:
 #### *I'm unable to bind to port 443 with unprivileged user*
 
 - Add value to /etc/sysctl.conf and reload daemon
-> `net.ipv4.ip_unprivileged_port_start=443`
-> `sysctl --system`
+    > `net.ipv4.ip_unprivileged_port_start=443`
+    > `sysctl --system`
 
 #### *I'm using a SSH solution that doesn't directly create a user.slice with a login*
 
 - Enable service persistance after logout
-> `loginctl enable-linger plextrac`
+    > `loginctl enable-linger plextrac`
 
 #### *I can't execute out of `/tmp` folder*
+
 - Download PlexTrac Manager Utility with wget using this command:
-> `wget -O ~/plextrac -q https://github.com/PlexTrac/plextrac-manager-util/releases/latest/download/plextrac && sudo chmod a+x ~/plextrac && sudo bash ~/plextrac initialize -v -c podman`
+    > `wget -O ~/plextrac -q https://github.com/PlexTrac/plextrac-manager-util/releases/latest/download/plextrac && sudo chmod a+x ~/plextrac && sudo bash ~/plextrac initialize -v -c podman`
 
 #### My containers don't start after rebooting the host VM
 
--- a/README.md
+++ b/README.md
@@ -64,6 +64,12 @@ Depending on your configuration, you may need to solve the following issues:
     > `net.ipv4.ip_unprivileged_port_start=443`
     > `sysctl --system`
 
+#### *I'm unable to use Let's Encrypt for SSL Certificate*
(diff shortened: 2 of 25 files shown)


mmorgan-pt (approved):
This all looks good, especially given the amount of testing that went into this.
Great work sir!

PR #121, opened 2024-03-21
Update version to 0.5.9

StevenPaugh:
Forgot to bump version again... Moves to 0.5.9 as it should be.


mmorgan-pt (approved):
Sorry I missed this too

PR #124, opened 2024-04-17
Update: Remove UID Dependencies

Michael-Burke:
Goal:
Update old logic to remove strict need to run as user `1337` as this can cause collisions with user provisioning and JIT solutions for the host OS

Task List:
- [x] Adjusted user requirement to require 'plextrac' user regardless of uid
- [x] Adjusted user creation to remove static assignment of the uid allowing for creation of an open UID on initialize
- [x] Create ability to ASSIGN uid on running `plextrac initialize`
- [x] Test permissions with volume binds / mounts
- [x] Test Backup and Restore processes
- [x] Test backwards compatibility with ownership by `1337`
- [x] QA full spectrum testing to catch permission / usage issues

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.6.0
+current_version = 0.6.1
 
 commit = True
 tag = True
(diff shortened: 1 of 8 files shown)


StevenPaugh (approved):
All changes look good. Upon a full/beta release of this I will test this out in my test instance to be sure it works as expected.

PR #125, opened 2024-04-18
Bugfix: Fix for Defunct Node processes

Michael-Burke:
This should fix the defunct node process issue as reported.

- Added init to two services (notification-engine, notification-sender)


mmorgan-pt (approved):
💪

PR #126, opened 2024-04-19
Add missing env var to notification-sender service

rkeeler:
Fixes a bug that is preventing emails from sending due to the notification-sender service not being able to make requests to our api because of a missing auth token.

Note: This is the same env var that the notification-engine service already has, but the notification-sender service needs it as well now.

### DIFF
--- a/static/docker-compose.yml
+++ b/static/docker-compose.yml
@@ -184,6 +184,7 @@ services:
       REDIS_CONNECTION_STRING: "${REDIS_CONNECTION_STRING:-redis}"
       REDIS_PASSWORD: "${REDIS_PASSWORD:?err}"
       serviceConfig: ${serviceConfig:-}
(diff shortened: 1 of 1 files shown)


static/docker-compose.yml
```diff
@@ -184,6 +184,7 @@ services:
       REDIS_CONNECTION_STRING: "${REDIS_CONNECTION_STRING:-redis}"
       REDIS_PASSWORD: "${REDIS_PASSWORD:?err}"
       serviceConfig: ${serviceConfig:-}
```
darius-plextrac, line 186: env var lines up with config 👍 

did we manage to add some alerting around the 401? That seems like it should scream at us.

darius-plextrac (approved):

static/docker-compose.yml
rkeeler: 👍 Great callout. Looking into that separately.

Michael-Burke (approved):

PR #128, opened 2024-04-19
Feature: Lock Update / Start If broken ETL

Michael-Burke:
- Adding locking mechanism to update and start to abort if present
- Added podman support for ETL locking
- Removed mod_rollout functionality

### DIFF
--- a/src/_docker_manager.sh
+++ b/src/_docker_manager.sh
@@ -30,6 +30,7 @@ function container_client() {
 }
 
 function image_version_check() {
+  die "Depricated: image_version_check"
   if [ $IMAGE_PRECHECK == true ]
     then
       IMAGE_CHANGED=true
@@ -87,14 +88,12 @@ docker.io/plextrac/plextracnginx:${UPGRADE_STRATEGY:-stable}"
 function pull_docker_images() {
   info "Pulling updated docker images"
   IMAGE_PRECHECK=true
-  image_version_check
   if tty -s; then
     ARGS=''
   else
     ARGS='-q'
   fi
   compose_client pull ${ARGS:-}
-  image_version_check
   log "Done."
 }
 
--- a/src/_docker_manager.sh
+++ b/src/_docker_manager.sh
@@ -30,7 +30,7 @@ function container_client() {
 }
 
 function image_version_check() {
-  die "Depricated: image_version_check"
+  die "Deprecated: image_version_check"
   if [ $IMAGE_PRECHECK == true ]
     then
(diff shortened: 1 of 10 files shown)


src/_docker_manager.sh
```diff
@@ -30,6 +30,7 @@ function container_client() {
 }
 
 function image_version_check() {
+  die "Depricated: image_version_check"
```
mmorgan-pt, line 33:
```suggestion
  die "Deprecated: image_version_check"
```

src/_manage_postgres.sh
```diff
@@ -136,3 +136,65 @@ fi
...
+
+function etl_failure() {
+  error "One or more ETLs are in an unhealthy status."
+  LOCK_UPDATES=true
+  LOCK_VERSION=${running_ver:-"failed"}
+  sed -i "/^LOCK_VERSION/s/=.*$/=${LOCK_VERSION}/" "${PLEXTRAC_HOME}/.env"
+  sed -i '/^LOCK_UPDATES/s/=.*$/=true/' "${PLEXTRAC_HOME}/.env"
+  sed -i '/^UPGRADE_STRATEGY/s/=.*$/=NULL/' "${PLEXTRAC_HOME}/.env"
+  
+  if [ "${LOCK_UPDATES:-false}" == "true" ]; then
```
mmorgan-pt, line 197: Why the check if it will always be true based on the initialization of the variable above?

src/_rollout.sh
```diff
@@ -25,7 +25,8 @@ scale() {
   compose_client up --detach --scale $service=$replicas --no-recreate "$service"
 }
 
-mod_rollout() {
+depricated_rollout() {
```
mmorgan-pt, line 28:
```suggestion
deprecated_rollout() {
```

src/_rollout.sh
```diff
@@ -25,7 +25,8 @@ scale() {
...
+  die "Depricated: mod_rollout"
```
mmorgan-pt, line 29:
```suggestion
  die "Deprecated: mod_rollout"
```

mmorgan-pt (changes requested):
An optimization question and couple spelling corrections

src/_manage_postgres.sh
Michael-Burke: Good catch -- fixed

mmorgan-pt (approved):
Tested and confirmed this works in a test environment for ubuntu 22.04

PR #131, opened 2024-04-29
Bugfix: REGEX correction on `plextrac stop` command

Michael-Burke:
- Fix REGEX and use of `docker` on the plextrac stop command


StevenPaugh (approved):
:shipit:

PR #132, opened 2024-04-29
Feature: Updated podman image paths to be dynamic

Michael-Burke:
- Added mechanism to use a custom registry with Podman


mmorgan-pt (approved):
Changes all look reasonable

PR #136, opened 2024-05-21
Added new function to control CKE migration with cli flag

Michael-Burke:
- Added new CLI flag `-cke | --cke-migration` to be used when updating to enable CKE

The system would mean enabling the environment variable, adding the service declarations into docker-compose, adding the new dockerhub token, and then running `plextrac update -v -cke`


mmorgan-pt (approved):
Tested and confirmed this all works as expected

PR #137, opened 2024-05-23
Feature: Added support for Air-gapped Mode

Michael-Burke:
- Basic Support to add environment var to toggle AIRGAPPED mode on
- This will prevent the manager util from updating, and the app from attempting to pull images. Docker Compose may still try if it can't find the images locally.


mmorgan-pt (dismissed):
One small suggestion in the documentation for a typo. Otherwise looks great and tested well 👍

PR #138, opened 2024-05-28
Added version comparison for post 1.6 for etl check

Michael-Burke:
- This is due to the script timing out / breaking on version 1.57 specifically, but we don't need ETL locking for updates until 2.0


mmorgan-pt (approved):
I successfully installed to 1.57, updated to 1.58, then set it to stable and let it rip all the way to 2.5. I didn't encounter any issues along the way and it automatically handled the updates, so this looks good!

PR #139, opened 2024-05-28
Bugfix: Install with RTC Failing

Michael-Burke:
## Tasks
- Added time component to migration timeout
- Reworked mod_start to separate migration call into `podman_run_cb_migrations` and `run_cb_migrations` and seeded this into where `mod_start` is being called appropriately
- Added new env vars to to migration container.
- Updated to support Podman (experimental)
- Updated to support RTC specific Image Registry as needed
- Removed strict coupling of the `plextrac` username for the non-root user

## Testing

**Status: In-Progress** ✅✅✅✅✅

- [x] Ubuntu
- [x] RHEL
- [x] CentOS
- [x] Debian
- [x] RHEL w/ Podman


StevenPaugh (approved):
Nice work dude. Bash lvl 100 💯

mmorgan-pt (changes requested):
only found a couple tweaks, otherwise looks great and validated functionality in an ubuntu environment

mmorgan-pt (approved):
Looks good!

PR #142, opened 2024-06-17
Bugfix: `PLEXTRAC_HOME` and Migration Container Names

Michael-Burke:
- Fixed container naming on pulling logs
- Adjusted container name expected when running couchbase-mitgrations container


mmorgan-pt (approved):
Tested changes and this looks good now!

PR #145, opened 2024-06-20
Bugfix: Removal of deprecated SMTP values

Michael-Burke:
- Removal of two unneeded values related to SMTP and using TLS/SSL with it


m4l1c3 (dismissed):
Based on syncing with Alex regarding settings/issues and a resolution which spurred this PR this appears correct.

PR #149, opened 2024-07-16
Bugfix: Suppress warning on normal updates

Michael-Burke:
- Adjusted the flag for self upgrades on update to correctly evaluate if SKIP_SELF_UPGRADE is set


StevenPaugh (approved):
LGTM

PR #152, opened 2024-09-12
Bugfix: Migration logs, `plextrac info` postgres bug, better messaging

Michael-Burke:
- Fixed warning about missing postgres container when running `plextrac info`
- Fixed `printf` usage when it's printing container logs to console and won't abort with 2.9 update
- Added "Update started" message after acknowledging warning so the user know's it didn't freeze
- Updated some Podman Documentation


mmorgan-pt (approved):
LGTM

PR #153, opened 2024-10-11
sp\logic for 2.10 Contiguous Updates

StevenPaugh:


mmorgan-pt (approved):
Tested following scenarios
1. pinned to specific version: no changes to upgrade strategy
2. set to stable with existing PLEXTRAC_MANAGED env var: no changes to upgrade strategy
3. stable without PLEXTRAC_MANAGED env var: changed upgrade strategy to 2.10 and did not encounter update issues

PR #155, opened 2024-10-11
Mm/semver bugfix

mmorgan-pt:


src/_version_check.sh
```diff
@@ -122,7 +123,8 @@ function version_check() {
         # Remove the running version from the Upgrade path
```
mrala, line 123: Why do we need to remove the running version for the array of `upstream_tags`?

src/_version_check.sh
```diff
@@ -141,7 +143,9 @@ function version_check() {
         # This grabs the first element in the version sorted list which should always be the highest version available on DockerHub; this should match stable's version"
         if [[ -n "${upstream_tags[*]}" ]]; then
```
mrala, line 144: If the goal is to get the latest semver version from docker tags, I feel like a lot of this code is redundant and can be replaced with:

```
latest_ver=$(printf '%s\n' "${upstream_tags[@]}" | sort -V | tail -1)
```

mrala (approved):

PR #157, opened 2024-11-06
SO-256: Ensure correct uploads volume ownership

darius-plextrac:
Adds an update check to ensure ownership of nodes in the `uploads` volume correctly and uniformly reflect the `plextrac` user's uid:gid. Applies to upcoming versions of the product, no-op in current and historical versions.


StevenPaugh (approved):
LGTM

PR #161, opened 2024-11-14
Changes Podman PG startup image

### LINKED ISSUE #160: Update default value for the PODMAN_PG_IMAGE env var
This line here:
https://github.com/PlexTrac/plextrac-manager-util/blob/main/src/_podman.sh#L156

The default should probably be `plextracpostgres:stable`.

---
We did see a customer instance fail to upgrade successfully from v2.7 to v2.8 because the pgvector migration mysteriously couldn't find the `vector` Type.
Customer issue tracked in Jira: https://plextrac.atlassian.net/browse/ES1-1090

Root cause was that the customer environment was still using the `postgres:14-alpine` docker image because they're using Podman and running PlexTrac in "airgap" mode.  The `PODMAN_PG_IMAGE` env var was not explicitly set in the customer environment and so the default value was being used, which does not include the `pgvector` Postgres extension

StevenPaugh:
resolves #160

### DIFF
--- a/src/_podman.sh
+++ b/src/_podman.sh
@@ -282,7 +282,7 @@ function podman_pull_images() {
 
   declare -A service_images
   PODMAN_CB_IMAGE="${PODMAN_CB_IMAGE:-docker.io/plextrac/plextracdb:7.2.0}"
-  PODMAN_PG_IMAGE="${PODMAN_PG_IMAGE:-docker.io/postgres:14-alpine}"
+  PODMAN_PG_IMAGE="${PODMAN_PG_IMAGE:-docker.io/plextrac/plextracpostgres:stable}"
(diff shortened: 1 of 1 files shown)


evanhsu (commented):
I can confirm that these changes reflect the fix that we made manually to resolve the customer issue.

Admittedly, I'm not super-familiar with the manager-util logic around Podman support, but I'm wondering if this PR also needs to address this other reference to the `postgres:14-alpine` docker image:
https://github.com/PlexTrac/plextrac-manager-util/blob/main/src/_docker_manager.sh#L44

...or if that's unrelated.

There are also 2 references to the `postgres:14-alpine` image in the airgapped readme file: https://github.com/PlexTrac/plextrac-manager-util/blob/main/docs/airgapped.md?plain=1#L23-L25

PR #167, opened 2024-12-02
minio service definiions updated

StevenPaugh:
rebound minio ports to localhost


mmorgan-pt (approved):
Tested and validated this change

PR #173, opened 2025-02-03
removed the conditional around check

mmorgan-pt:


src/_stop.sh
```diff
@@ -7,9 +7,8 @@ function mod_stop() {
   title "Attempting to gracefully stop PlexTrac..."
   debug "Stopping API Services..."
 
-  # Before stopping, check if stable is being used and the current image tag for stable is correct
-  if [[ "$UPGRADE_STRATEGY" == "stable" ]]; then
-    debug "Running stable, validating the expected version against current running version"
+  # Before stopping, check if the current image tag matches the image defined by compose files
```
mrala, line 10: Should lines 11-29 need to be de-indented since the outer if loop was removed?

```suggestion
  # Before stopping, check if the current image tag matches the image defined by compose files
  debug "Validating the expected version against current running version"
  running_backend_version="$(for i in $(compose_client ps plextracapi -q); do docker container inspect "$i" --format json | jq -r '(.[].Config.Labels | ."org.opencontainers.image.version")'; done | sort -u)"
  running_frontend_version="$(for i in $(compose_client ps plextracnginx -q); do  docker container inspect "$i" --format json | jq -r '(.[].Config.Labels | ."org.opencontainers.image.version")'; done | sort -u)"
  expected_backend_tag="$(compose_client config | grep image | grep plextracapi | head -n 1 | awk '{print $2}')"
  expected_frontend_tag="$(compose_client config | grep image | grep plextracnginx | head -n 1 | awk '{print $2}')"
  expected_backend_version="$(docker image inspect $expected_backend_tag --format json | jq -r '(.[].Config.Labels | ."org.opencontainers.image.version")')"
  expected_frontend_version="$(docker image inspect $expected_frontend_tag --format json | jq -r '(.[].Config.Labels | ."org.opencontainers.image.version")')"


  if [[ "$running_backend_version" != "$expected_backend_version" ]]; then
    error "The running backend version ${running_backend_version} does not match the expected version (${expected_backend_version})"
    error "During a system reboot or shutdown, the docker engine normally handles this gracefully and automatically, so using 'plextrac stop' may be unnecessary"
    die "Since 'plextrac stop' runs a docker compose down, we cannot guarantee a 'plextrac start' will bring up the correct version. Please change UPGRADE_STRATEGY to the current running version ${running_backend_version} or run an update first"
  fi
  if [[ "$running_frontend_version" != "$expected_frontend_version" ]]; then
    error "The running frontend version (${running_frontend_version}) does not match the expected version (${expected_frontend_version})"
    error "During a system reboot or shutdown, the docker engine normally handles this gracefully and automatically, so using 'plextrac stop' may be unnecessary"
    die "Since 'plextrac stop' runs a docker compose down, we cannot guarantee a 'plextrac start' will bring up the correct version. Please change UPGRADE_STRATEGY to the current running version ${running_frontend_version} or run an update first"
  fi
```

mrala (changes requested):
Should lines 11-29 need to be de-indented since the outer if loop was removed?

mrala (approved):

PR #177, opened 2025-03-20
Mm/automated enhancements

mmorgan-pt:
This change breaks the docker compose files into 3 separate files depending on the desired configs.

* If the upgrade strategy is set to a value before 2.0, then it keeps the current compose file
* If the strategy is above 2, or set to stable, it grabs a newer compose file that sets the CouchbaseDB to the 7.2.0 image
* If the CKEDITOR_SERVER_LICENSE variable is set, then the related env vars are set and the compose file with the ckeditor RTC services is used

How this has been tested:
* Tested in a VM with an installation for 1.59, then set to stable. It upgraded couchbase and the system as expected.
* Set a CKEDITOR_SERVER_LICENSE_KEY and verified the env vars created as expected, along with the compose file changes
* Tested with similar compose file as cloud hosted and verified no collisions

Further testing to do:
* others should review code and test steps listed above
* a beta version should be released and used with a cloud hosted instance to ensure no issues arise


EramusBlack1:
yes this is definitely going the right way

src/_configure_plextrac.sh
```diff
@@ -75,10 +75,15 @@ CLOUD_STORAGE_ACCESS_KEY=${CLOUD_STORAGE_ACCESS_KEY:-`generateSecret 20`}
 CLOUD_STORAGE_SECRET_KEY=${CLOUD_STORAGE_SECRET_KEY:-`generateSecret`}
 CLOUD_STORAGE_ENDPOINT=${CLOUD_STORAGE_ENDPOINT:-"minio"}
 CLOUD_STORAGE_SSL=${CLOUD_STORAGE_SSL:-"false"}
+CKEDITOR_SERVER_LICENSE_KEY=${CKEDITOR_SERVER_LICENSE_KEY:-}
+`setCKEMigrateConfig`
+CKEDITOR_ENABLE_METRIC_LOGS=false
+CKEDITOR_LOG_LEVEL=40
```
Michael-Burke, line 81: Is this the value we want on by default in production? I haven't checked but want to validate this to be sure.

src/_docker_manager.sh
```diff
@@ -99,7 +99,7 @@ function pull_docker_images() {
 
 function composeConfigNeedsUpdated() {
   info "Checking for pending changes to docker-compose.yml"
-  decodedComposeFile=$(base64 -d <<<$DOCKER_COMPOSE_ENCODED)
+  #decodedComposeFile=$(base64 -d <<<$DOCKER_COMPOSE_ENCODED)
```
Michael-Burke, line 102: Should probably replace this with the updated DOCKER_COMPOSE_ENCODED value since this function handles drift and keeps this compose file from manual changes

Michael-Burke (commented):
Overall very nice! I'd like to see the bits I mention below cleaned up, but otherwise I think its good.

src/_configure_plextrac.sh
mmorgan-pt: Yes, that came from the on-prem docs that we provide customers

src/_docker_manager.sh
mmorgan-pt: Ahhh, I remember now. This is handled correctly, but rather than setting the var decodedComposeFile here, it's set in the _configure_plextrac.sh, so when it does the check here, it catches any pending changes. I'll just remove this line now so as to not cause any confusion

src/_configure_plextrac.sh
```diff
@@ -131,6 +136,17 @@ function setDefaultSecrets() {
   export IFS=$OLDIFS
 }
 
+function setCKEMigrateConfig() {
+  if [ -z ${CKEDITOR_SERVER_LICENSE_KEY:-} ]; then
+    echo CKEDITOR_MIGRATE=false
+  else
+    echo CKEDITOR_MIGRATE=true
+    echo IMAGE_REGISTRY=${IMAGE_REGISTRY:-docker.cke-cs.com}
+    echo IMAGE_REGISTRY_PASS=${IMAGE_REGISTRY_PASS:-}
```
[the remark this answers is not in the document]
mmorgan-pt: Totally fair, I'll update that

PR #182, opened 2025-03-28
adds the integration worker service

mmorgan-pt:


mrala (approved):
Looks good! Do we need to make any changes in salt?

PR #193, opened 2025-04-21
Adds volume mount that integration-worker needs when moving ctem artifactors to reporting

runia1:
## Description

The integration-worker service needs the volume mount for the uploads folder.  We store ctem artifacts there.  The integration-worker only reads these artifacts when moving data from ctem into a report but without the volume mounted, it logs a bunch of warnings trying to look up the artifact files, and the data gets moved into reporting without artifacts, which isn't ideal.

## Testing

We should be able to test this by manually applying this 2 line change to a docker compose instance via the override file to verify this is a safe change.

## Backwards compatibility

This is an additive change only, older versions of the integration-worker should not require the uploads volume mount, and if given the mount should not cause any issue.


jfjermestad-plextrac (approved):

mmorgan-pt (dismissed):
LGTM

mmorgan-pt (dismissed):

mmorgan-pt (approved):

PR #194, opened 2025-04-24
ES1-1281: default to empty for env vars that may not exist for on-prem customers

jordanreedie-plextrac:
### What

Provide an explicit (empty) default value for env vars that are not required (and that we do not expect on-prem customers to populate). This should have no impact to functionality, since docker was already populating the value with a blank string when it the variable is not set. This just tells docker not to make a stink about it in the event that it's not there.

### Why
This silences the "WARN" messages that docker outputs when it can't find an env var that has been specified without a default. Example:
```
WARN[0000] The "JAEGER_ENDPOINT" variable is not set. Defaulting to a blank string.  
```
This has resulted in a few support tickets being opened by confused customers.

#### To reproduce the issue: 
Run a plextrac update without those env vars set. See warnings that look like the following for each of the modified env vars:
```
WARN[0000] The "JAEGER_ENDPOINT" variable is not set. Defaulting to a blank string.  
```

#### To test the fix
Apply the changes in this branch and run a plextrac update without those env vars set. You should _not_ see the WARN messages described above


mmorgan-pt (approved):
Tested and confirmed this works. Nice work!

PR #196, opened 2025-05-30
Mm/drop database

mmorgan-pt:


src/_restore.sh
```diff
@@ -103,6 +103,9 @@ function restore_doPostgresRestore() {
       # tear down the existing postgres container, including the related volumes
       compose_client down $postgresComposeService --volumes
 
+      # stop the rest of the app to avoid issues with writes coming into the fresh database before a restore
+      compose_client down
```
rkeeler, line 107: Would it be better to call `mod_stop`?

mmorgan-pt: Yes, but I hadn't figured out why the stop function was working yet. I just pushed up a fix for that and it should be good now

PR #215, opened 2026-03-16
Do not delete export_* files from uploads folder

runia1:
These files are report export files. We've historically never allowed retrieval outside of the export endpoint where they're generated but we are going to begin to allow historically generated reports to be downloaded and tracked in the database. This means we can no longer delete them on a cron interval.

All generated exports should begin with the word `export_`


StevenPaugh:
LGTM

PR #220, opened 2026-04-22
Add FORCE_KEYCLOAK_MIGRATION_RUN variable to docker-compose

mmorgan-pt:


rkeeler (approved):
👍  Correct env var applied to the correct container

PR #221, opened 2026-07-07
Feat: Deprecate `cbbackup` for `cbbackupmgr` for Couchbase

Michael-Burke:
## Summary
Switches Couchbase backup/restore to `cbbackupmgr` by default, with a new `--legacy` flag to fall back to the deprecated `cbbackup`/`cbrestore` tools. Also fixes two latent shell-scripting bugs uncovered while validating this, and adds application-version + type labeling to backup archives.

**Why:** `cbbackup` silently abandons DCP streams that stall >30s and still exits 0, causing undetected partial backups. `cbbackupmgr` is the actively-maintained replacement and doesn't have this bug.

**Changes:**
- `src/plextrac`: added `--legacy` flag (sets `LEGACY_BACKUP=true`)
- `src/_backup.sh`: `backup_fullCouchbaseBackup` now dispatches to `cbbackupmgr` (default) or `cbbackup` (`--legacy`). Legacy path validates transfer completeness (transferred vs. estimated msg count, stall detection) instead of trusting exit code alone.
- `src/_restore.sh`: same split for `restore_doCouchbaseRestore`.
- `src/_manage_couchbase.sh`: `CB_BACKUP_USER` now provisioned with `admin` instead of `bucket_full_access` - `cbbackupmgr` requires roles (`data_backup`/`backup_admin`) that don't exist on Community Edition, so `admin` is the only role that works. Added `test_couchbase_backup_role` so `plextrac autofix`/`plextrac check` can still catch and fix this manually. Also added `ensure_couchbase_backup_user_role`, called from `plextrac update` (`_update.sh`), so already-provisioned instances get upgraded automatically on their next update without requiring a manual `autofix` - idempotent, a no-op once fixed.
- `insufficient_credentials` failures now hint to run `plextrac autofix` instead of just printing the raw Couchbase error.
- **Bug fix:** "find latest backup" lookups switched from `ls -dc1` (ctime) to `ls -dt1` (mtime). The `chown -R` permissions fix that runs beforehand bumps ctime on every file, which can make `ls -c` fall back to alphabetical order once ctimes cluster together - silently restoring the wrong file. Confirmed live: the new `cbbackupmgr-archive-*` naming sorted incorrectly under the old lookup.
- **Bug fix:** `tar -tzf <large-archive> | head -n1` under `pipefail` can crash the whole script - `head` exits early, `tar` gets `SIGPIPE` (141) trying to keep writing, and `set -e` treats that as fatal. Confirmed live (a real restore crashed with an opaque stack trace). Fixed with `set +o pipefail` / `set -o pipefail` around the four affected lookups. `k3s_restore.sh`'s identical-looking pattern was never at risk since it runs in a nested subshell that doesn't inherit `pipefail`.
- **New:** couchbase/postgres/uploads archives now get a type + version suffix, e.g. `2026-07-08T163256z-uploads-v3.2.5.tar.gz`, applied unconditionally regardless of `--legacy`. Added `get_plextrac_version()` (reuses the label lookup already in `_version_check.sh`).
- Fixed `_restore.sh`'s archive-directory lookup to read the tarball's own contents instead of assuming the outer filename matches it (`basename -s .tar.gz`) - that assumption broke once the version suffix was added.
- Suppressed the `redis-cli` password warning (`--no-auth-warning`) on the three license-cache-clear calls in `_restore.sh`/`k3s_restore.sh`.

**Security note:** this is a real privilege increase for the backup user (bucket-scoped → full cluster admin), unavoidable on Community Edition. `--legacy` remains available if bucket-scoped least-privilege is required.

**Note:** `--legacy` is required on restore if the backup being restored was itself taken with `--legacy` (archive formats aren't cross-compatible).

**Validation:** tested restore-side discovery against synthetic archives covering all three naming generations (pre-version, version-only, version+type) for couchbase (cbbackupmgr and legacy layouts), postgres, and uploads. All 15 cases pass - discovery is entirely tar-content-based, so the outer filename never mattered.

## Test plan
- [ ] Run `plextrac autofix` on an existing instance and confirm `CB_BACKUP_USER` gets upgraded to `admin`
- [ ] Run `plextrac update` on an instance with a stale `bucket_full_access` `CB_BACKUP_USER` and confirm it gets upgraded to `admin` automatically, with no separate `autofix` needed
- [ ] Run `plextrac update` again afterward and confirm it's a no-op for this check (no re-prompt, no re-fix)
- [ ] Manually reset `CB_BACKUP_USER` back to `bucket_full_access` and confirm `plextrac autofix` still catches and re-fixes it
- [ ] Run `plextrac backup` and confirm couchbase/postgres/uploads archives are named `<timestamp>-<type>-vX.Y.Z.tar.gz`
- [ ] Run `plextrac backup --legacy` and confirm old tool behavior + validation still works, and naming is unchanged
- [ ] Run `plextrac restore` against a cbbackupmgr-taken backup, a legacy-taken backup (`--legacy`), and an old pre-version backup
- [ ] Confirm mismatched tool/archive combo fails with the hint message, not a confusing error
- [ ] Confirm a stale bucket-scoped `CB_BACKUP_USER` fails `plextrac backup` with the `plextrac autofix` hint
- [ ] With mixed legacy/cbbackupmgr backups present, confirm restore picks the actually-newest file
- [ ] Confirm restore no longer crashes with a SIGPIPE trace on a large couchbase archive
- [ ] Confirm the redis license-cache-clear step no longer prints the password warning


mmorgan-pt (dismissed):
Given level of testing, this should be ready to go

rohitkumar1369 (approved):

